PR #25, opened 2015-06-29
Pluggable Authentication via Passport

katexyu:
We wanted to use sqlpad internally at Gumroad, but we wanted to use one of our existing identity solutions instead of username/password. To that extent, we integrated with Passport.js (a commonly used npm module that provides a multitude of authentication strategies). You can now:

  * turn off the username-password strategy (by setting the `DISABLE_USERPASS_AUTH` environment variable,
  * turn on the Google OAuth strategy (by setting `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET`, and `PUBLIC_URL`.

Let us know if there's anything we can do to make this change more consumable/welcome :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "connect-flash": "^0.1.1",
     "cookie-session": "1.0.x",
     "ejs": "1.0.x",
+    "errorhandler": "^1.4.0",
     "express": "4.x.x",
     "json2csv": "2.2.x",
     "keymaster": "^1.6.2",
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "bcrypt-nodejs": "0.0.3",
     "body-parser": "1.2.x",
     "cookie-parser": "1.1.x",
+    "connect-flash": "^0.1.1",
     "cookie-session": "1.0.x",
     "ejs": "1.0.x",
     "express": "4.x.x",
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "mysql": "2.x.x",
     "nedb": "1.x.x",
     "passport": "^0.2.2",
+    "passport-google-oauth2": "^0.1.6",
     "passport-local": "^1.0.0",
     "pg": "^4.1.0",
     "pg-cursor": "^1.0.0",
--- a/package.json
+++ b/package.json
@@ -43,6 +43,8 @@
     "mssql": "1.x.x",
     "mysql": "2.x.x",
     "nedb": "1.x.x",
+    "passport": "^0.2.2",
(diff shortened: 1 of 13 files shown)


rickbergfalk:
This looks amazing! I've heard of passport but never got around to looking into it that closely. The fact that you've cleaned up my hacky auth with it and added google auth as an option is huge. Totally welcome change. Thanks so much! :)

And to top it off Gumroad is using SqlPad? very cool. 

I'll run some tests with existing sqlpad instances tonight and if all is good publish a new version to npm... 

Thanks again!

thatha:
We're also doing some testing on our end—I think we'll have to fix a couple of things.
Thanks for being so responsive!

rickbergfalk:
I did some quick testing, and the local auth strategy appears to work just fine with existing SqlPad instance. The only minor bug I found was always redirecting to the connections page if an admin, since it wasn't waiting for the results of the connections findOne query to come back. 

How are things looking on your end? 

I'll hold off publishing to npm until I get the go-ahead from your side of things :)

katexyu:
Oh, you're right; we accidentally put the check outside the callback. Another thing is `req.logout()` in the `/signout` function doesn't seem to be clearing the session properly, so that will be replaced with `req.session = null;`, which seems to work. We'll open another pull request with any changes once we've consolidated them on our end. Thanks!

thatha:
Aside from what Kate mentioned above, all good for us.
Thank you for being so responsive, it's very encouraging.

rickbergfalk:
Great! Thanks again to you both and you're welcome! I'll try to have the other requests merged in and published to npm within the next few days.

katexyu:
Ok great, let us know if there's anything you need from us or if you see anything that you'd like to see changed, etc. Thanks!

rickbergfalk:
Will do! Sorry for the delay I've been pretty busy with the holiday weekend and all. I'll have things in  tomorrow evening at the latest :)

rickbergfalk:
Not sure if you saw already but all these pull requests have been published to npm under version 1.8.0. Thanks again - these are some great adds!

katexyu:
Yeah I saw, thanks and you're welcome!

rickbergfalk:
Also @katexyu - since you've taken the time to make some great additions, and it sounds like Gumroad is getting some use out of SqlPad, I've gone ahead and made you a collaborator on the repository. No pressure to do anything with the power or anything... I'm not really sure what power/benefit there is beyond push access. I just like to give some ownership of the project to anyone that puts in the time to make it better :)

And if there's any reason you don't want to be an official collaborator that's cool too.

PR #33, opened 2015-08-21
Implemented forever to use the application in production

valterjrdev:
Added script for server initialization environment
Added to package.json libraries: forever, sys and child-proc
Added to README.MD instructions for server startup

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,11 +57,18 @@ Optionally install watchify to automatically browserify your client-side scripts
 npm install watchify -g
 ```
 
-To start SqlPad on port 3000  with datafiles in ./db run 
+To start the development environment SqlPad on port 3000  with datafiles in ./db run 
 
 ```sh
 npm start
 ```
+
+To start SqlPad in a production environment using forever
+```sh
+npm start --env=prod --db=./db --port=8080
+```
+Instructions on how to use the forever are in https://github.com/foreverjs/forever
+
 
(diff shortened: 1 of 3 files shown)


rickbergfalk:
Thanks @valterjrdev! I skimmed over the additions and they look nice! I won't have any time today to merge in and publish to npm, but I will try to get to it this weekend.

If I understand things correctly, this is to enable running SqlPad for production use directly from the git repository, instead of installing and running via npm install -g right?

valterjrdev:
Before it was only possible to start the server to work in the development environment, the case goes to the terminal server finished.
Now it is possible to leave the server running in and out of the machine passing the argument --env = prod, indicating that is an instance of production and running the library forever.

rickbergfalk:
This has been published to npm

rickbergfalk:
@valterjrdev Just a heads up - I'm going to be reverting this pull request. I'm not sure the workflow is going to work out. By running the commands as a child process none of the console.log output is shown (which is useful for debugging). I also noticed forever is included in the package.json, but the way its being used still requires it to be installed globally separately by the user. 

I think using forever is a good idea though. Instead would it be alright if you just ran ```forever start ./server.js``` directly? (or is there any reason you couldn't run it directly?)

Sorry for not digging into this more in depth before I merged the pull request - Despite reverting this I appreciate the effort you put into this :).

PR #34, opened 2015-08-21
Prevent DB password from being sent to the frontend.

katexyu:
Right now the application sends the DB password along with the connection username to the admin interface, which seems unnecessary and less secure. For example, if your deployment of SqlPad isn't using HTTPS for some reason, someone can snoop on your network and obtain your DB credentials.

The tradeoff is that admins won't be able to grab the password directly from the interface.

### DIFF
--- a/routes/connections.js
+++ b/routes/connections.js
@@ -36,7 +36,7 @@ module.exports = function (app) {
                 connection = {};
             } else {
                 connection.username = decipher(connection.username);
-                connection.password = decipher(connection.password);
+                connection.password = "";
             }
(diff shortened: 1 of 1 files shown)


dcelasun:
Sounds perfectly reasonable. Don't know what @rickbergfalk thinks, but I'd say you can go ahead and merge.

rickbergfalk:
I agree - I think its a good idea to keep it from the client. Thanks @katexyu

rickbergfalk:
This has been published to npm

PR #54, opened 2015-10-06
feat(embeddable chart): for any query, add a page with just the chart

nbrustein:
I want to be able to create dashboards on my company's internal wiki with multiple charts.  This PR makes it possible for me to do that using iframes pointing at sqlpad charts.

This PR adds a button under 'Save Chart Image' with the label 'Link to Chart'.  That links to the page for the current query with `format=chart` in the querystring.  The page that is rendered shows only the chart for the query, without the header or any of the editing functions.


rickbergfalk:
Thanks @nbrustein! I've been a bit under the weather lately but I hope to get to these this weekend.

rickbergfalk:
I merged these in for now. still not up to reviewing it and testing it - hope to get to that tomorrow. Excited about this functionality though! Was kind of thinking about this sort of thing already and you were already on it. very cool :)

rickbergfalk:
@nbrustein finally got to checking it out - really nice! your pull requests have been published to npm under version 1.10.0.

Would you like to have commit access to the repo? Not entirely sure what the benefits are but I like to do that for anyone that makes a significant contribution :)

Something to think about - I don't believe the embed links will work if the person viewing the link isn't logged in to SqlPad. I'm not sure about that though. I'd be willing to consider opening up the access for chart/table only renders, but it might be worth opening an issue for an having a discussion about. Maybe its something people opt in to? 

Thanks again!

nbrustein:
Hey Rick,

Sure, I would accept commit access.  I promise not to abuse the power.

And yeah, you're right about needing to be logged in to sqlpad.  It's not a
great user experience in my case, because I have a dashboard with 6
different charts pulled from sqlpad, and if you're not logged in then they
all show the login form.  Even worse, each frame is too small to really fit
the login form.  to make this a little bit better, on the page containing
the iframes I have text reminding people that they need to login to sqlpad
first.  It's not great, but it's an internal tool, so it's good enough.

For my page, assuming we stick with multiple iframes on the page, I guess
the best reasonable thing would be:

   1. Each iframe has a link that says "login"
   2. Clicking any of those links opens a pop-up window with a sqlpad login
   form.
   3. Once the form is filled out, each iframe refreshes (One way to
   implement this is that each iframe is polling for a change to a `loggedIn`
   flag stored in local storage, and once it changes to true, all the iframes
   reload themselves.)

That would all be a bit hackish, but the whole thing of using iframes is a
bit hackish.  It's practical, though, as far as I'm concerned, so might be
worthwhile doing.

As for allowing non-logged-in people to see charts, that's probably a good
feature, though not one we would need over here (at pedago
<http://pedago.com/#smartly>), since we wouldn't want to open up this data
to the world.

Thanks for getting that stuff merged in.  And thanks for this project.
It's been really helpful for us.

 - Nate

PR #58, opened 2015-10-25
The user must be registered to compare the password hash.

dudemelo:
Trying to sign in  with an authorized account but not yet signed, the server will crash, because we are passing an empty password to the bcrypt compare.


dcelasun:
Nice catch! Merged.

PR #71, opened 2015-12-22
Adding Env variables to port and passphrase

shukydvir:
For use in Docker image when you run the image.

### DIFF
--- a/lib/add-cli-config-to-app.js
+++ b/lib/add-cli-config-to-app.js
@@ -10,9 +10,9 @@ module.exports = function (app) {
     // $HOME/.sqlpad/db
     //var configFile = path.join(userHome, '.sqlpad/config');
     var config = rc('sqlpad', {
-        port: 80,
+        port: process.env.SQLPAD_PORT || 80,
         dbPath: path.join(userHome, "sqlpad/db"),
-        passphrase: "At least the sensitive bits won't be plain text?"
+        passphrase: process.env.SQLPAD_PASSPHRASE || "At least the sensitive bits won't be plain text?"
     });
     
     // support --db and --dir for database directories


dcelasun:
I think it's time we have:

- A common format (like SQLPAD_FOO, SQLPAD_BAR etc.) for all environment variables (e.g passport stuff)
- A config system that supports both key value pairs set from the UI *and* via environment variables.

shukydvir:
Sure I agree.

Do you have any format in mind ?

The reason I added it now since I built a Docker image and I want it to be flexible passing things into the container via ENV vars.

Thx

dcelasun:
I'm thinking we can map `SP_FOO` to just `foo` in the config system (case insensitive). We already have a few environment variables from #25, but I think we can move them to `SP_` as well and still allow the old versions for BC purposes.

@rickbergfalk what do you think?

rickbergfalk:
I'm definitely on board with cleaning up and formalizing the config stuff.

SqlPad is using rc (https://www.npmjs.com/package/rc), which will already bring in environment variables beginning with the app name... although I don't believe we are taking advantage of it at all.

@shukydvir SqlPad may already honor the environment variables in the pull request... If not, they might need to be lowercase? (if they do need to be lowercase, that's a bummer and I'm fine with explicitly adding support via the pull request)

@dcelasun - is there a reason you'd prefer environment variable prefix to be SP_ over SQLPAD_?

rickbergfalk:
Did some quick tests with the rc package. Capitalization does matter unfortunately.

So as is, I think sqlpad should automatically pick up environment variables sqlpad_port and sqlpad_passphrase but not SQLPAD_PORT and SQLPAD_PASSPHRASE.

I'm in favor of supporting the all-caps env variables though.

Thoughts?

dcelasun:
SP_ seems more managable to me if you have lots of env variables, but it's purely a cosmetic concern.

On capitalization, if rc is case-sensitive, I think we should just pick one (lower/uppercase) and stick with it. I'm personally fine with either.

rickbergfalk:
I like the SQLPAD_ prefix for clarity and all caps because... its just what I've traditionally seen more? 

I will set aside some time today to publish an update to npm. Thanks @shukydvir!

rickbergfalk:
This has been published to npm under version 1.13

PR #74, opened 2015-12-30
Fix out-of-date underscore and lodash dependency and use latest lodash

dlabey:
.sortByOrder was not apart of the lodash in there


dlabey:
Latest is broken and I cannot chart. Trying to figure out which version of lodash will work best.

PR #76, opened 2016-01-03
Add New Feature  [BeautiSQL]:Which Use to format the SQL

kira8565:
1、add new feature name [BeautiSQL] to make The SQL More BeautiFul
2、add new Method <code>setEditorText</code>  for aceSqlEditor to Set The SQL after format

### DIFF
--- a/client-js/component-ace-sql-editor.js
+++ b/client-js/component-ace-sql-editor.js
@@ -3,24 +3,28 @@ var ace = require('ace');
 
 module.exports = function (id) {
     var me = this;
-    
+
     id = id || "ace-editor";
     var editor = ace.edit(id);
-    
-    if (editor) { 
+
+    if (editor) {
         //editor.setTheme("ace/theme/monokai");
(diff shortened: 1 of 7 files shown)


rickbergfalk:
Hi @kira8565 

I really appreciate the enthusiasm you've put into this, but unfortunately I don't think it is a good fit for SqlPad. 

SQL formatting can be very preferential and difficult to get right. I ran a simple query through the jQuery plugin the formattor module was based off of, and found that comments can sometimes impact the query after formatting. For example: 

http://zachofalltrades.github.io/jquery.format/demo/index.html

```sql
SELECT 
    column1, -- this is a comment
    column2,
    column3
FROM 
    sometable
```	

becomes...

```sql
SELECT column1,
         -- this is a comment column2,
         column3
FROM sometable
```

I appreciate the effort you've applied here though. If you'd like to contribute to SqlPad, I encourage you to try tackling one of the open issues. :)

PR #81, opened 2016-01-17
update dependencies and remove old files

rickbergfalk:
Some dependencies not updated. Run npm outdated to see which. Either
these libraries require a larger effort to  update or they are early
major versions


rickbergfalk:
wait forgot to include and up-to-date browserified bundle :(

PR #84, opened 2016-01-18
Add support for binding to a specific IP address

dcelasun:
Node's HTTP server binds to 0.0.0.0 by default, but there are situations where you want to bind to a specific address (multiple interfaces on a single host, SSL without SNI etc.). This commit adds support for that while preserving BC.

@rickbergfalk I'd like to merge this to master soon, but I'm open to changes/feedback in the meantime.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next Release
 - Many client-side and server-side dependencies updated
+- Add ability to bind to a specific IP address via the --ip flag or the SQLPAD_IP environment variable
 
 ## 1.14.0
 - Add ability to turn off date localization (add config item "localize" set to "false")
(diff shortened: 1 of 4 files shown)


rickbergfalk:
This looks good to me - feel free to merge to master whenever you feel comfortable with it :)

Can binding IP addresses be used to restrict which IPs can access the service? I'm thinking of postgres config, where you have to explicitly open it up to listen to 0.0.0.0 if you want it to be open to the entire world...

dcelasun:
Merged!

> Can binding IP addresses be used to restrict which IPs can access the service? I'm thinking of postgres config, where you have to explicitly open it up to listen to 0.0.0.0 if you want it to be open to the entire world...

Exactly. I'd like to bind SqlPad to `127.0.0.1` so it won't be reachable from outside the host and allow access only through an nginx reverse proxy.

Issue #90, opened 2016-02-08
Dockerize


tclass:
Hi, 
would be nice to have a Dockerfile direct in this repo and maybe and image on hub.docker.com

tclass:
wow thx, looks great!

rickbergfalk:
As I commented in #91 I think this is a great idea, I'm just not familiar with docker. 

I've created a SqlPad organization on hub.docker.com and will work towards getting an image published hopefully later tonight. (https://hub.docker.com/u/sqlpad/)

Any assistance welcomed :)

marcmillien:
Sure :).
My docker username is as github `marcmillien` ;).

rickbergfalk:
@marcmillien I've added you to the contributors team on the sqlpad org at hub.docker.com. I wasn't sure how I should manage the team structure... any ideas/suggestions let me know. 

I'm enjoying learning about docker, but honestly I wouldn't mind handing off docker image management too if that's something you'd be interested in owning :)

marcmillien:
While I can, I'm ready to help in this :).

You have to create some repository, I visibly can't do that :).
https://hub.docker.com/u/sqlpad/dashboard/

Maybe we should follow you versioning for the docker tags.
I mean we would have:
```
sqlpad/sqlpad:latest
sqlpad/sqlpad:1.15.0
...
```
I'm going to create a `Makefile` to help us pushing this to the docker hub.

rickbergfalk:
Thanks again @marcmillien! This is great. I'm in agreement with the versioning strategy too. 

I created a repository and added you as an admin. Hopefully you should be able to push the image now?

Otherwise I'll probably get around to this later this weekend. Thanks for all the assistance on this.

marcmillien:
Done :)
https://hub.docker.com/r/sqlpad/sqlpad/tags/

I think we have to change the `docker-compose.yml` for a developper perspective.
I'll work on it when I have time to.

tclass:
guess I close the issue than :D thx everyone

PR #91, opened 2016-02-08
Adding a docker build

marcmillien:
Ref.: https://github.com/rickbergfalk/sqlpad/issues/90

Just test it using: `docker-compose up`

You'll need docker & docker-compose ;).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,6 @@
+web:
+  build: docker
+  environment:
+    SQL_PAD_PASSWORD: somepassword
+  ports:
+    - "3000:3000"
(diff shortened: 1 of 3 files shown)


dcelasun:
Seems fine, but I'm not at all experienced with docker, so I'll let someone else decide on this :)

marcmillien:
Next step would be create some organisation on hub.docker.com then push the resulting image :).
This container will probably need improvements, but that's a first step ^^.

rickbergfalk:
Thanks @marcmillien. I'm not experienced with docker either, but I like this being an option. Docker has been one of those technologies on my backlog to learn...

@marcmillien would you like to be part of the SqlPad org on docker? No pressure if you'd rather not.

Otherwise I'll try and learn this stuff later tonight :)

PR #92, opened 2016-02-10
[docker] Adding a Makefile for pushing to docker hub registry

marcmillien:
Ref.: https://github.com/rickbergfalk/sqlpad/issues/90#issuecomment-182150570

Just running `make` will create then upload the new docker image to docker hub for those having the right to push to sqlpad organisation.

To specify the version to build, just add the version as an argument to `make`, for example:
```
make SQLPAD_VERSION=1.0.0
```


rickbergfalk:
Thanks @marcmillien!

PR #108, opened 2016-06-29
break out inlined schema sql, add material views (pg)

rickbergfalk:


rickbergfalk:
Thanks for reporting this @dashwin! I initially had that but it errors out when developing. I'll investigate and have a fix out shortly

rickbergfalk:
fix is published under 1.17.1

PR #119, opened 2016-09-23
Enhanced docker ecosystem

marcmillien:
This PR:
- replaces the docker entrypoint by a better one:
  - we offer the possibility for users to choose their `sqlpad` arguments
  - for example if you want to set `--passphrase` or `--base-url` then you can simply run docker with this arguments
- removes the `Dockerfile.tpl` for a real `Dockerfile`, it's more standard and it opens the door for futur usage of automated builds, more details at https://hub.docker.com/add/automated-build/sqlpad/ (it needs some work to be done today like having npm build sript in the repository, I don't know how you build it @rickbergfalk ?)
- upgrades to nodejs 4.5.0
- directly build docker from the official docker nodejs image
- creates a better docker-compose.yml for developpers working in the project


rickbergfalk:
Thanks @marcmillien this looks really nice! Automated builds would be great. I'll look into what needs to be done to set that up. 

Some background for building sqlpad - 

When you clone the github repository and run `npm install` npm will install both the regular dependencies and dev-dependencies listed in the package.json file. 

When an end user installs sqlpad via npm via `npm install sqlpad -g`, npm will only download the regular dependencies. 

For v2, the command `npm run build` will run the build.js file, which will build the front-end javascript. Most of these front-end dependencies are listed as dev-dependencies. The bundled file is published to npm, so end users do not need to download all the dev-dependencies in the package.json.

marcmillien:
Thanks for the details :), I'm going to test it ^^.

PR #124, opened 2016-10-24
Add connector for REST API queries

csavelief:
Hi,

The goal of this PR is to allow people to easily connect to REST API backends. I am not sure it will be of interest to you but it serves us well!

Best,

Cyrille SAVELIEF


rickbergfalk:
Hi @csavelief looks interesting. 

To help me understand better, is this for utilizing your own sql execution web service that is exposed as a restful api? (I could see how that might be useful to support a database that doesn't have great node.js support)

csavelief:
Exactly, although we do not use your editor to send SQL queries but a custom DSL. Luckily for us, your project does not try to mess with the query typed in the editor :-) We are running an Accumulo cluster behind the REST API.

rickbergfalk:
I appreciate the pull request @csavelief. After giving it some thought I'm going to close this. I think this is a nice idea, but requiring code changes to configure it wouldn't be beneficial for most sqlpad users. If this were somehow configurable via the ui I think it'd be an interesting addition.

Issue #127, opened 2016-11-07
SMTP over different ports


jameswinegar:
Going to look at adding SMTP configuration with different ports, mainly 587

rickbergfalk:
Hi @jameswinegar - I'd gladly accept a pull request if you take this on. 

Some help to get you started: you'll want to add an SMTP Port config item here: https://github.com/rickbergfalk/sqlpad/blob/master/resources/config-items.toml

Once that is in place, you can reference the new value wherever email is used, like here: https://github.com/rickbergfalk/sqlpad/blob/master/routes/users.js#L68

Adding a config item will add it to the UI as long as all the information is there. 

Let me know if you need any help - this is a pretty low-key project :)

jameswinegar:
Got this working. Just working on the encryption piece when sending over 587.

`{ Error: 140737231827904:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:794:
 code: 'ECONNECTION', command: 'CONN' }`

PR #128, opened 2016-11-08
Add SMTP Port

jameswinegar:
Added SMTP Port as a config option, setup tls cipher for use with office 365. Tested with gmail and office 365. #127


rickbergfalk:
Looks great! - thanks @jameswinegar!

Issue #126, opened 2016-11-07
HTTPS support


jameswinegar:
I've never written anything in Node so I was wondering if this is as easy to implement as I think. Would I just need to edit the `server.js` file such that if a key is in config or passed in environment then to require https and run that way?

rickbergfalk:
I've never implemented https via express directly, but I have seen some code snippets, and I do believe its as easy as passing in the required config. (a quick google led me to this post http://blog.mgechev.com/2014/02/19/create-https-tls-ssl-application-with-express-nodejs/)

If you'd like to take this on as well, you would approach the config in a similar way you'd approach the config for the SMTP port, by defining a config item in [config-items.toml](https://github.com/rickbergfalk/sqlpad/blob/master/resources/config-items.toml)

Since this is used for creating the expressjs http server, you probably want to use interface = "env". And maybe 2 settings, one for key file path and another for cert file path? 

The few times I've set this up, I've use nginx in front of sqlpad, running sqlpad on an alternate port (like 3000) and then forwarding the a subdomain to the sqlpad port, with nginx handling the HTTPS piece. 

Definitely welcome a pull request if you get this figured out though! Feel free to reach out if you need anything.

jameswinegar:
My use case requires both changes! So hopefully I can make them work tomorrow sometime. Read through the code base today to make sure I understood it. 

I want to make the http vs https robust enough so people don't have to create keys and blah if they don't need it.

rickbergfalk:
Hi @jameswinegar - this was included in the v2.1.0 publish to npm. thanks! Hope you've enjoyed node so far :)

PR #129, opened 2016-11-08
Sqlpad https

jameswinegar:
I don't think this is the right way to handle the logic for deciding whether to do HTTPS or not, but it's a step in the right direction. Thoughts on what to do better here?

#126

### DIFF
--- a/server.js
+++ b/server.js
@@ -134,14 +134,16 @@ app.use(function (req, res, next) {
 ============================================================================= */
 
 var fs = require('fs');
-// determine if key pair for cert exists
-var keyPairExists;
-if (fs.accessSync(KEY_PATH,'r') == null && fs.accessSync(CERT_PATH,'r') == null){
-    keyPairExists = true
-}
 
-if (keyPairExists) { //https with http redirect
+// determine if key pair exists for certs
+if (KEY_PATH.replace(/\s/g,"") != "" && CERT_PATH.replace(/\s/g,"") != "") { //https only
     console.log('Launching server with SSL')
+    fs.access(KEY_PATH, fs.constants.R_OK, (err) => {
+        console.log(err ? 'No key found or is not readable' : 'SSL key readable');
+    });
+    fs.access(CERT_PATH, fs.constants.R_OK, (err) => {
+        console.log(err ? 'No cert found or is not readable' : 'SSL certificate readable');
+    });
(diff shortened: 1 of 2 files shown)


rickbergfalk:
Hi James! I think this is a good start.

I can see how some of this could be tricky - there's a lot of traditionally async operations here reading files and testing ports, which can quickly get complicated. 

Luckily since this is core config needed for the operation of the server, I'm okay if sync operations are used as possible and if errors are thrown to crash the application if an invalid config is present.

I'll add some comments to the code with some tips :)

server.js
```diff
@@ -127,18 +132,58 @@ app.use(function (req, res, next) {
...
-  }
-  http.createServer(app).listen(_port, IP, function () {
-    console.log('\nWelcome to ' + app.locals.title + '!. Visit http://' + (IP === '0.0.0.0' ? 'localhost' : IP) + ':' + _port + BASE_URL + ' to get started')
-  })
-})
+
+var fs = require('fs');
+
+// determine if key pair exists for certs
+if (KEY_PATH.replace(/\s/g,"") != "" && CERT_PATH.replace(/\s/g,"") != "") { //https only
```
rickbergfalk: instead of using replace to clean whitespace, you could do something like ```KEY_PATH.trim()```, which will remove surrounding whitespace and leave any value. Even simpler, js treats an empty string as a falsey, and not empty as truthy, so you could naively do

```js
if (KEY_PATH && CERT_PATH) { 
  // do https setup
}
```

Of course this doesn't work if the user sets one of these values to a white-space padded string, so maybe you'll still want to use .trim()

rickbergfalk:
Added some comments - If you make any changes and push to your branch it should update this pull request.

Another minor thing - if you could adjust the formatting that'd be great. I've been using the standard tool to fix/enforce a style preset: http://standardjs.com/ It's an opinionated style, but not configurable so its less hassle.

If you install the tool with ```npm install standard -g``` you should be able to format the js from the command line automatically with ```standard --fix```

Hope all these comments are helpful aren't too overwhelming! This is a really great start, especially for just starting with Node. 👍 

Thanks again for taking the time to contribute! :)

marcmillien:
I'm asking myself, why not using nginx in front for handling ssl ? ^^
This is what we are currently doing here ;).

rickbergfalk:
Thanks @jameswinegar!

PR #130, opened 2016-11-15
Support non English characters when downloading files

askluyao:
If the HTTP header contains non English characters, an error will be thrown about "invalid characters" and the application will be terminated. This problem is probably caused by enhanced check by Node.js

Related issues: 
https://github.com/indexzero/http-server/issues/244
https://github.com/npm/npm/issues/11515

So I use encodeURIComponent to encode the file name. Files with Chinese characters could be downloaded now.

Thanks.


rickbergfalk:
Thanks @askluyao!

Issue #94, opened 2016-03-03
Swapping back and forth between SQL and Vis resets legend filtering


ybroze:
If one toggles certain series on and off in the plot, toggling to the SQL view and back removes the previous selections. Would be nice if they were preserved.

rickbergfalk:
This has been addressed in the v2 branch :) I hope to have the beta published to npm in a few weeks.

PR #136, opened 2016-11-21
Render chart lazily

vweevers:
We found Taucharts has problems with big datasets, freezing or even crashing the browser. And because sqlpad renders the Vis tab even when it's not visible - at the same time as rendering the result table - this meant we weren't able to view query results either.

With this PR, the chart is only rendered (once) if the Vis tab is active. Like before, the chart is kept around to preserve its internal state, so the user can switch tabs back and forth.

When the user re-runs the query, the chart is destroyed. Rather than rendering when the query results change, we destroy when the results are *about to* change. So at that point we're back to the initial state and can render lazily again.

Moving the first `renderChart` call to `componentDidUpdate`, i.e. when the DOM has been updated to make the chart `div` visible, also means you no longer need the `setTimeout` hack.

Additionally, the Visualize button is disabled when there's nothing to chart.

### DIFF
--- a/client-js/QueryEditor.js
+++ b/client-js/QueryEditor.js
@@ -380,25 +380,19 @@ var QueryEditor = React.createClass({
     })
   },
   sqlpadTauChart: undefined,
+  hasRows: function () {
+    var queryResult = this.state.queryResult
+    return !!(queryResult && queryResult.rows && queryResult.rows.length)
+  },
+  isChartable: function () {
+    var pending = this.state.isRunning || this.state.queryError
+    return !pending && this.state.activeTabKey === 'vis' && this.hasRows()
+  },
   onVisualizeClick: function (e) {
     this.sqlpadTauChart.renderChart(true)
   },
   onTabSelect: function (tabkey) {
     this.setState({activeTabKey: tabkey})
-    var renderChartIfVisible = () => {
-      var chartEl = document.getElementById('chart')
-      if (chartEl.clientHeight > 0) {
-        try {
-          this.sqlpadTauChart.renderChart()
-        } catch (e) {
-          console.log('tauchart rendering failed')
-          console.log(e)
-        }
-      } else {
-        setTimeout(renderChartIfVisible, 20)
-      }
-    }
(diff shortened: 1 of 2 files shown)


rickbergfalk:
Thanks for your work on this @vweevers! This is really nice. Testing this out and I can already tell the initial query result grid is much quicker to load. Your changes are a much more elegant wrapper around taucharts too :) 

The only thing I'm a bit hesitant on is the chart being destroyed between query runs. There was previously an issue about keeping the filters between switching tabs (#94) which this does, but in addressing that I also attempted to keep them around executing a new query. 

I don't know if this is that important though. It doesn't impact my workflows and I'd rather prefer a snappier render and less crashing if I end up querying a larger dataset. TauCharts filters are also kind of this tacked on thing -- They aren't preserved when saving a query.

What do you think? 

@ybroze any thoughts on this?

vweevers:
Are you saying that TauCharts preserves filters if one calls `chart.setData(newResult)`? I would not expect it to, thinking filters are tied to specific rows of the previous result.

rickbergfalk:
it does... although testing it further I'm noticing some odd behaviors. 

For example, in a multi-line chart, if you hide a line and re-run the query, the line will still be hidden. If query run 2 does not have the data for that series, and query run 3 does again, that line will still be hidden. Oddly if you hide a series and then run a query that adds more series, the additional series will be colored all the same color line. 

Another weird behavior I've discovered with time-series line charts and the date filter - if you filter down to a time segment, then run the query on a larger time span, the date filter widget will be stuck with the first query's date range. The full range of query 2 won't be available until you fully render the chart again. 

I'm starting to think that this is probably a minor edge case, and if someone is relying on this feature they might be seeing odd behaviors as is. 

I personally think performance should be the winning priority here :)

vweevers:
I guess for TauCharts by itself, there are plenty use cases for preserving filters. Under condition that the data source is controlled or at least predictable. In sqlpad that's not true, it's at the mercy of the user, so preserving filters doesn't make much sense here.

> I personally think performance should be the winning priority here :)

Agreed, and thanks for the quick merges! 👍

Issue #131, opened 2016-11-15
TypeError: Cannot read property 'table_type' of undefined


santoshsahoo:
After migrating from 1x:

The object browser fails to load with error:

browserified.js:64 TypeError: Cannot read property 'table_type' of undefined(anonymous function) @ browserified.js:64

rickbergfalk:
Hi @santoshsahoo - Could you try forcing a refresh on the schema browser by clicking the refresh icon? the object formatting behind the scene changed between v1 and v2, and I think your sqlpad instance might be picking up the cached schema (though I thought this was getting cleared between v1 and v2, so not sure what is going on here)

Let me know if that fixes your issue..

vweevers:
I had the same issue. For some reason, the migration did not run at all (but no error was logged either, so I didn't find out this was the issue until later). Because of that, we lost the admin roles, and got the schema error described here. Only when we copied the old database again and manually removed the cache (`rm -rf db/cache db/cache.db`) before anything else, did the migration run.

rickbergfalk:
I haven't had time to look too far into this, but I recall testing it a few times to ensure the migrations ran successfully.

The db migrations are run after the nedb database loads, but it doesn't prevent the rest of sqlpad from doing its thing, so I'm wondering if that could cause a problem...

Or maybe this is a permissions/OS/node version thing? I'd think an error would at least be logged if that were the case though.

vweevers:
I can't replicate it anymore. My best guess is that the cache datastore was broken somehow, failed to load, and then the migrations didn't run because of the if-statement in [lib/db.js#L48](https://github.com/rickbergfalk/sqlpad/blob/b443628aa3c14828194852f2c36fd2fd712cdbf3/lib/db.js#L48):

```
if (db.loadedDbs.length === db.instances.length) {
      onAllDatabasesLoaded()
}
```

vweevers:
Update: it might (also) be an issue with nodemon and npm's shell quoting. I noticed that a change to `db/schemaVersion.json` - which *should* be ignored - triggers a restart. This is problematic because this file changes in every step of the migration. So the next steps (as well as other stuff happening at the same time) might not get a chance to safely finish.

vweevers:
To make the ignore patterns work cross-platform, we can place them in a `nodemon.json`:

```json
{
  "ignore": ["public/*", "client-js/*", "db/*"]
}
```

vweevers:
[...]

PR #150, opened 2017-01-03
Strict startup order

vweevers:
The startup order becomes:

- load databases
- migrate schema
- create indexes
- save configuration
- create admin
- set auto compaction
- start HTTP server (last, so we don't process requests until the db is ready)

All steps are run serially to avoid race issues and other conflicts. If any of the steps fail, the app will throw an error.


vweevers:
In part to address #131.

rickbergfalk:
Thanks @vweevers! Do you run sqlpad via the npm install or from your a local repo? 

Just wondering if you'd prefer these changes published to npm right away or if it could wait a bit...

vweevers:
@rickbergfalk from a private fork. So it doesn't matter to me. This is just to give something back ❤️

PR #152, opened 2017-01-11
update sql schema to reflect the latest changes in Crate

mikethebeer:
Column names in `information_schema` have changed since Crate 1.0.0
https://github.com/crate/crate/blob/master/CHANGES.txt#L129

### DIFF
--- a/resources/schema-crate.sql
+++ b/resources/schema-crate.sql
@@ -1,13 +1,13 @@
 select 
     'Tables' as table_type,
     'YES' as is_nullable, 
-    tables.schema_name as table_schema, 
+    tables.table_schema as table_schema, 
     tables.table_name as table_name, 
     column_name, 
     data_type 
 from 
     information_schema.tables, information_schema.columns 
 where  
-    tables.schema_name not in ('information_schema') 
-    and columns.schema_name = tables.schema_name 
+    tables.table_schema not in ('information_schema') 
+    and columns.table_schema = tables.table_schema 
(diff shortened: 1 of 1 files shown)


rickbergfalk:
Thanks @mikethebeer - I want to merge but want to think further about how to handle this with sqlpad versioning. This would be a breaking change of sorts for any users of crate < 1.0.0 right? 

Any idea if it is common for crate users to keep their crate installs up to date? 

To avoid the break, I'm thinking I might add a major version dropdown field to the crate connection form (with options like 0.x and 1.x) and adjust the schema sql accordingly...

mikethebeer:
I like your idea with the version dropdown field for different Crate versions. However, changing the column name of the Information Schema was an exceptional case that occurred because of our first Major `1.0` version which is already declared "stable". Usually the CrateDB community is quite flexibel and fast in doing version updates. Your solution with the dropdown 0.x and 1.x is good IMHO.

rickbergfalk:
@mikethebeer this should be fixed in the latest beta published to npm. Would you be willing to test this out? 

Instead of having users select a crate version in the connection screen, sqlpad will first try to run the query for crate v1. if that returns an error, it'll run the schema query for v0. 

You can check out the beta by running `npm install -g sqlpad@beta`

mikethebeer:
@rickbergfalk thanks for fixing it. I'm fine with the solution you made - Already checked it and works well. 👍 thanks

PR #153, opened 2017-01-11
Add PostgreSQL certificate support

nikicat:
May be it's not so good implementation, but it works (tests are missed)


lib/run-query.js
```diff
@@ -154,7 +155,10 @@ clients.postgres = function (query, connection, queryResult, callback) {
     password: connection.password,
     database: connection.database,
     host: connection.host,
-    ssl: connection.postgresSsl
+    ssl: connection.postgresSsl ? {
```
rickbergfalk-uptycs: I'm not entirely certain, but I think we still need to support ssl: true here. Setting to true without passing in any cert or key will use ssl but not alter any ssl parameters... (I think anyways - I've only set this once a long time ago when querying heroku postgres)

rickbergfalk-uptycs:
thanks @nikicat! I'll happily merge - just one small fix to consider :)

nikicat:
Ok, @rickbergfalk-uptycs, I'll fix it

rickbergfalk:
@nikicat I'm doing some refactoring with the way the connections UI is done... merging this into my dev branch so we don't lose out on your work :)

PR #169, opened 2017-03-10
Support for postgre SSL CA

johicks:
Based on @nikicat 's work, added support for a CA field.

I was able to successfully connect to my database after this change, however, the 'Test' button still shows the connection fails even though it works once saved.


rickbergfalk:
Thanks @johicks! I was going to recommend looking for the changes in the earlier PR but it looks like you already took care of it. I'll take a look into what's going on with the test button.

johicks:
@rickbergfalk I think the test is actually working fine, I mistakenly typoed a field!

PR #180, opened 2017-05-05
Option to connect to postgres through SOCKS proxy

brysgo:
Don't merge this yet, I'm going to test it tomorrow.

Just wondering if you would be interested in something like this. I only did it because we have to deal with a database behind a SOCKS proxy at work and the node pg client is the *only* way I have been able to connect to it so far, and this seems to be the only pg admin tool that uses node.

### DIFF
--- a/client-js/connections/ConnectionForm.js
+++ b/client-js/connections/ConnectionForm.js
@@ -68,6 +68,31 @@ const fields = {
     formType: TEXT,
     label: 'Database CA Path'
   },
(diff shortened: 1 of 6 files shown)


brysgo:
Okay, you can merge it if you like it, it works.

rickbergfalk:
Thanks @brysgo! This looks great. I've never had to deal with a SOCKS proxy yet but I'll gladly bring this in :)

PR #185, opened 2017-05-28
Add systemd socket activation support

epeli:
I have a situation where sqlpad is used only randomly by one person so to save server resources I want to spawn sqlpad only when it is actually being used.

Systemd can itself listen to the socket and start the application on demand when the socket is accessed. The socket itself is passed to the process as file descriptor 3 when the `LISTEN_FDS` env is defined by systemd.

Systemd configuration should be something  like following:

`/etc/systemd/system/sqlpad.service`
```
[Unit]
Description=sqlpad
After=network-online.target
Wants=network-online.target systemd-networkd-wait-online.service

[Service]
User=sqlpad
Group=sqlpad
Environment=NODE_ENV=production
WorkingDirectory=/apps/sqlpad/sqlpad
# Shutdown automatically after 8h
RuntimeMaxSec=28800
ExecStart=node server.js --port systemd --ip 127.0.0.1 --dir /apps/sqlpad/data --passphrase secret
```

and `/etc/systemd/system/sqlpad.socket`
```
[Socket]
ListenStream=127.0.0.1:3000 
[Install]
WantedBy=sockets.target
```

Reload the systemd daemon

    systemctl daemon-reload

Start the socket and enable it on boot

    systemctl start sqlpad.socket
    systemctl enable sqlpad.socket

Now the sqlpad node process is started only when the port 3000 is being accessed. This could be improved by implementing automatic inactivity shutdown in sqlpad but for now I just use the `RuntimeMaxSec` option to shutdown it after 8 hours.

### DIFF
--- a/docs/_pages/installation-and-administration.md
+++ b/docs/_pages/installation-and-administration.md
@@ -135,4 +135,6 @@ A SQLPad query can be posted to a Slack webhook when saved. To enable, create a 
 
 An entire domain can be whitelisted for username administration by setting enviornment variable ```WHITELISTED_DOMAINS```
 
+### Systemd socket activation
 
+To use systemd socket activation set port to magic value ```systemd```. For more information see [this pull request](https://github.com/rickbergfalk/sqlpad/pull/185).
--- a/server.js
+++ b/server.js
@@ -136,6 +136,26 @@ if (fs.existsSync(htmlPath)) {
   console.error('If not running in dev mode please report this issue.\n')
 }
 
+// When --port=systemd is passed we will assume that sqlpad is launched as a
+// systemd service with socket activation. Systemd passes the socket as file
+// descriptor 3. As sqlpad listens to only one port no further action is
+// required.
+//
+// More info
+//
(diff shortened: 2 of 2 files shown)


rickbergfalk:
Thanks @epeli I'll gladly pull this in. 

Before I do though I just want to make sure I understand... if sqlpad is installed on a system using socket activation for a different application, would its regular startup be prevented since LISTEN_FDS is defined on the system? 

I'm wondering if there needs to be an additional config flag passed to sqlpad so it knows it should be listening for socket activation instead of starting regularly (something like `node server.js --ip 127.0.0.1 --dir /apps/sqlpad/data --passphrase secret --systemd-socket`

epeli:
The `LISTEN_FDS ` env is only defined when sqlpad is started as a systemd service with socket activation. 

But I do agree that it's better to only explicitly enable this feature. Instead of creating new flag I added special case for `--port=systemd` option. I also refactored this PR a bit to support socket activation with https too. Added a bit of documation too.

rickbergfalk:
Awesome - thanks!

PR #187, opened 2017-05-29
Add normal port binding fallback

epeli:
Found a nasty edge case with the `--port systemd` style. If the systemd socket unit is stopped and the service is started as a normal systemd service without the socket unit it will crash because it assumes it will get a port but it does not. So sqlpad must implement a fallback when the socket is not found.

So the systemd configuration must be like this

`/etc/systemd/system/sqlpad.service`
```
[Unit]
Description=sqlpad
After=network-online.target
Wants=network-online.target systemd-networkd-wait-online.service

[Service]
User=sqlpad
Group=sqlpad
Environment=NODE_ENV=production
WorkingDirectory=/apps/sqlpad/sqlpad
# Shutdown automatically after 8h
RuntimeMaxSec=28800
ExecStart=node server.js --systemd-socket --port 3000 --ip 127.0.0.1 --dir /apps/sqlpad/data --passphrase secret
```

and `/etc/systemd/system/sqlpad.socket`
```
[Socket]
ListenStream=127.0.0.1:3000 
[Install]
WantedBy=sockets.target
```

Maybe I should write a wiki page about this configuration? I have a feeling that this is a bit obscure setup for the sqlpad home page :)

### DIFF
--- a/docs/_pages/installation-and-administration.md
+++ b/docs/_pages/installation-and-administration.md
@@ -137,4 +137,4 @@ An entire domain can be whitelisted for username administration by setting envio
 
 ### Systemd socket activation
 
-To use systemd socket activation set port to magic value ```systemd```. For more information see [this pull request](https://github.com/rickbergfalk/sqlpad/pull/185).
+To use systemd socket activation add ```--systemd-socket``` flag. For more information see [this pull request](https://github.com/rickbergfalk/sqlpad/pull/185).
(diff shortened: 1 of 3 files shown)


rickbergfalk:
Thanks @epeli! A wiki would be good I think - I was thinking about reworking the main project page anyways... 

I created an `Administration and Configuration` page in the wiki. Feel free to add notes. Repo settings say any user should have access - let me know if that isn't the case.

PR #230, opened 2017-10-06
Fix docker image entrypoint shebang

adrienbrault:


rickbergfalk:
Thanks for catching this!

PR #261, opened 2017-11-08
BigNumbers fixed

KochamCie:
BigNumbers fixed


KochamCie:
hi  .
my colleague found when a column defines to BigInt type,the results have been somekind scientific notation.like 6209255456651680769 to 6209255456651681000.
then i config the 'supportBigNumbers'

rickbergfalk:
Thanks @KochamCie!

PR #262, opened 2017-11-10
 enable drag to view column cell content

KochamCie:
enable drag to view column cell content

### DIFF
--- a/client-js/common/QueryResultDataTable.js
+++ b/client-js/common/QueryResultDataTable.js
@@ -6,8 +6,8 @@ import '../css/fixed-data-table.css'
 import './QueryResultDataTable.css'
 
 const renderValue = (input, fieldMeta) => {
-  if (input === null || input === undefined) {
-    return <em>null</em>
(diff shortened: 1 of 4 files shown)


KochamCie:
e'm'm
my friend found when the result cell content is large while the div element can't move to see detail.
even though I can change the column width .

so i play a trick =0=

rickbergfalk:
haha ok I'll try this out :)

Any idea why past commits already merged are showing up in your PR? Have you pulled latest from sqlpad master?

KochamCie:
I'm weird, too  ：（

rickbergfalk:
Sorry for the delay in getting back to you! I tried your branch out and it is indeed a creative solution :)

Unfortunately though it loses styling...

What do you think about having a button on each row to see a popup of all the rows results? That way long text could be viewed all at once?

KochamCie:
i just found a great bug in this solution.hah
so i'm going to close it
your idea was my first thought~  but my css skill is poor

PR #274, opened 2017-12-09
Use Tachyons for application styles

rickbergfalk:
This swaps out most the custom/application css for Tachyons classes. Eventually I'd like to replace bootstrap with basic components and Tachyons and simplify a lot of the UI to make it more utilitarian and basic. 

Tachyons brings a lot of benefits and cleans up the lack of consistent styling strategy, but it brings some drawbacks too (people can't customize by overriding css classes, now people need to learn tachyons)

I personally think the benefits Tachyons brings outweigh the drawbacks, and I partially think making it harder to customize/override css classes is a good thing (if people want configurable styles or change the way it looks, contribute!)

That said - I'm going to leave this open for a few days in case anyone out there wants to try to convince me to not do this :)


rickbergfalk:
Taking the lack of input as an ok to go ahead with this. If anyone out there is like what is going on with the CSS in this app and finds this PR, and wants to chat or raise an issue or anything really please do not hesitate to reach out (for this or anything really). 

Next styling updates will be to make SQLPad a bit more "boring" and utilitarian. I think the bootstrap list/card things are going to be replaced with tables. Full screen forms, side nav. 

Then eventually rip bootstrap out.

PR #296, opened 2018-02-02
Addition for SAP HANA

ccmehil:
Addition for SAP HANA as a database option with SQLPAD


rickbergfalk:
Thanks so much for this @ccmehil! Looks great. I should have this in the next release some time next week.

PR #353, opened 2018-04-21
Add tests and notes for HANA

rickbergfalk:
* Adds tests for hana
* Adds notes and TODOs for hana implementation
* Updates runQuery() rows array validation error message
* rearranges hana fields order


rickbergfalk:
FYI @ccmehil - Thanks for blogging about running HANA under docker!

PR #371, opened 2018-06-12
"SQLPad aims to a SQL" → "SQLPad aims to be a SQL"

Naereen:
Should I also make the same change [on the .html page in the doc](https://github.com/rickbergfalk/sqlpad/blob/f7b3856cc7935792006007ff6296107594657646/docs/index.html#L215) ?


rickbergfalk:
Thanks for catching this @Naereen! I'll re-run the build process and it'll update the pages

PR #386, opened 2018-10-01
Add column descriptions to schema sidebar for Postgres

greigwise:


server/drivers/postgres/index.js
```diff
@@ -29,7 +29,8 @@ const SCHEMA_SQL = `
     ns.nspname as table_schema, 
     cls.relname as table_name, 
     attr.attname as column_name,
-    trim(leading '_' from tp.typname) as data_type
+    trim(leading '_' from tp.typname) as data_type,
+    pg_catalog.col_description(attr.attrelid, attr.attnum) as col_desc
```
rickbergfalk: What do you think about changing the field name to either `description` or `column_description`? The latter is a bit long, but I'm just thinking about the consistency in variable names...

greigwise: column_description is fine I think.  I can do that.

client-js/queryEditor/SchemaSidebar.js
```diff
@@ -335,20 +336,38 @@ class SchemaInfoColumnItem extends React.Component {
...
-        >
-          {column_name}
-          <span className="silver"> ({data_type})</span>
-          {getCopyToClipboard()}
-        </span>
-      </li>
-    )
+    if (col_desc) {
+      return (
+        <li className="list">
```
rickbergfalk: Instead of having the full `<li>` node duplicated, could you refactor to put the description into a variable and then reference that in a single `<li>`?

perhaps something like 

```js
const description = col_desc && ` - ${col_desc}`;

return (
  <li>
    {/* all the other stuff */}
    <span className="silver"> ({data_type})</span>{description}{getCopyToClipboard()}
  </li>
)
```

greigwise:
OK.  I think that should address your request.

rickbergfalk:
Thanks @greigwise! Looks great

PR #415, opened 2019-03-15
Bring semicolons back!

rickbergfalk:
RIP meaningful file updated dates, git blame, etc. Sorry if you've been maintaining a fork of SQLPad (kinda). 

I ditched the semicolon in this project 2.5 years ago. I don't know that I had a preference for or against it, but instead I liked the standardjs tooling in that it handled formatting and eslint all together. It was opinionated in some areas, but not too much in others, and at the time it seemed like a good fit. 

Shortly after that decision I got a job as a JavaScript developer. There we eventually adopted Prettier when it became a thing. We used semicolons. 

Switching back and forth between codebases with and without semicolons has been a struggle to be honest. It always take a bit to adjust. 

At this point I don't care what the code format is, I'd just like it to be consistent.

And I figure it'll be the most consistent if I use the most popular formats (which I assume must be the prettier defaults?)

So semicolons are back.

### DIFF
--- a/client/src/AboutContent.js
+++ b/client/src/AboutContent.js
@@ -1,9 +1,9 @@
-import PropTypes from 'prop-types'
-import React from 'react'
+import PropTypes from 'prop-types';
+import React from 'react';
 
 class AboutContent extends React.Component {
   render() {
-    const { version } = this.props
+    const { version } = this.props;
     return (
       <div>
         <p>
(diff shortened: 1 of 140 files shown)


rickbergfalk:
I shoulda mentioned I might use the default quotes too at some point. We'll see...

PR #453, opened 2019-08-02
SAML integration

deemoowoor:
This is the code required to integrate SQLPad with SAML authentication. With correct configuration, which one can create using the guidance at http://www.passportjs.org/packages/passport-saml/, it's possible to connect this to a SAML authentication provider, such as AD FS, for example. 

This version maps users by their emails.


client/src/SignIn.js
```diff
@@ -93,11 +93,19 @@ function SignIn({ config, smtpConfigured, passport, refreshAppContext }) {
...
+  const samlForm = (
+    <div>
+      <a href={config.baseUrl + '/auth/saml'}>Sign in with ADFS</a>
+    </div>
+  );
+
   return (
     <div style={{ width: '300px', textAlign: 'center', margin: '100px auto' }}>
       <h1>SQLPad</h1>
+      {JSON.stringify(passport.strategies)}
```
rickbergfalk: I think this was maybe accidentally left from debugging?

package.json
```diff
@@ -24,5 +24,8 @@
   "scripts": {
     "fixlint": "npm run fixlint --prefix client && npm run fixlint --prefix server && prettier --write '**/*.js'",
     "lint": "npm run lint --prefix client && npm run lint --prefix server && prettier --check '**/*.js'"
+  },
+  "dependencies": {
+    "cipm": "^2.0.0"
```
rickbergfalk: Curious about cipm. Is this necessary?

server/middleware/must-be-authenticated.js
```diff
@@ -9,7 +9,7 @@ module.exports = function mustBeAuthenticated(req, res, next) {
     return next();
   }
   if (req.headers.authorization) {
-    return passport.authenticate('basic', { session: false })(req, res, next);
+    return passport.authenticate('basic', { failureRedirect: '/', failureFlash: true })(req, res, next);
```
rickbergfalk: Did this require changing? It has been a while but if I recall this was in place for HTTP basic auth to allow API access using username and password. I'm not sure the failureRedirect/failureFlash do anything here

client/src/SignIn.js
deemoowoor: Ahh, yes, overlooked it during cleanup.

package.json
deemoowoor: Nice catch. I'll check if it is at all needed. Could be one of my tweaking attempts while I was trying to make build work. 

Vanilla repo wouldn't build on my local machine, because apparently some of the dependencies in the ```package-lock.json``` could not be properly installed. I think one of the suggestions I found online referred to ```cipm``` to be used, so I added it as a dependency. But I don't think it helped.

After all, what actually fixed the build for me was removing package-lock.json altogether and re-generating it with ```npm install```.

server/middleware/must-be-authenticated.js
deemoowoor: I'll clean this up.

server/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "scripts": {
     "prepublishOnly": "cd .. && ./scripts/build.sh",
-    "start": "node-dev server.js --dir ../db --port 3010 --debug --base-url '/sqlpad'",
+    "start": "node-dev server.js --dir ../db --port 3010 --debug --base-url /sqlpad",
```
deemoowoor: Notice how base-url is not quoted? On Windows/MINGW at least, these single quotes were treated as part of the URL.

package.json
rickbergfalk: You can run `npm ci` in the directory of a package-lock.json to install those dependencies. There are 3 package-lock files in this project... `scripts/build.sh` installs all of them. I should probably add those separate steps in the README

deemoowoor: Well, ```scripts/build.sh``` wouldn't install them for some reason. Also, I seem to have to run it twice for it to do all the right things, so there's definitely something wrong with that. Or maybe it's just Windows...

server/package.json
rickbergfalk: I was going to ask about this. If I recall not having the single quote might cause problems on OS X. Not a huge deal.

package.json
deemoowoor: But yes, ```cipm``` is not needed here. Will remove.

rickbergfalk:
@deemoowoor it looks like cipm is still in the root package-lock.json file. 

Could you either revert the change to that file, or delete the root package-lock.json and recreate it by running `npm install` in root of the project?

rickbergfalk:
👌 Thanks again @deemoowoor!

rickbergfalk:
@deemoowoor Do you see any issue with updating passport-saml version 1.2.0? The major version difference indicates a breaking change. 

v1 was out at the time of this PR, so I was curious if there was something that didn't work with it?

I don't have a saml test setup handy...

rickbergfalk:
Release notes for convenience https://github.com/bergie/passport-saml/releases

deemoowoor:
Upon reading the release notes, I don't see any issues. 

Node.js v4 is a pretty old version (I'm using v10 in all my projects). It's fine to let it go now.

Signature validation behaviour change looks fine by me too. Sure it may break stuff, but it should, since it indicates incorrect and unsafe use. Not checking signatures when not provided was a serious security flaw, IMHO.

deemoowoor:
I could test it, but no earlier than next week. I'm planning to go live with my SQLPad project next week, actually, so I'll probably have a time window to test it out.

Issue #301, opened 2018-02-07
Custom authentication


AtomBaf:
Hi

For corporate use inside an intranet, I'd like to implement a custom authentication scheme.

Basically, the SQLPAD is located behind a reverse proxy that will handle the authentication, then proceed to SQLPAD with special HTTP headers (X-Forwarded-User, X-Forwarded-Email, ...).
SQLPAD should then just check the headers to know the currently connected user. (There is no need for a signin page)

I tried to implement a passport strategy with no success. The passport strategy is something like https://www.npmjs.com/package/passport-reverseproxy or https://www.npmjs.com/package/passport-trusted-header

So, could you help and implement this strategy ? or, could you add a wiki page and some APIs so that anyone can implement its own authentication scheme ? 

This would be much appreciated.
Many thanks for your great product

ccbourloud:
I have the same problem, it would be great to find a solution !

rickbergfalk:
Thanks for the kind words @AtomBaf! I don't have any documentation handy at the moment or any direction other than to take a look at the couple passport implementations that are in SQLPad now...

I'm on the verge of organizing the project directory structure so it makes a bit more sense. Once I have that out of the way I can loop back and give you some tips?

I could even get something started and stubbed out as an experimental thing... but no promises beyond that... though maybe this is something that is easily testable since it is just inspecting http headers?

I'll come back to this with more info, but it might be a week or two :)

fulldecent:
Here is code I am using for a reverse proxy in a PHP environment that already authenticates the signon. Provided for reference and as a test case.

```php
<?php
namespace PMTDashboards;

# Authenticate this request
define('REQUIRED_AUTHORIZATION', 'management');
require '../../config-production.php';
$configuration = Configuration::production();

// https://raw.githubusercontent.com/michaelfranzl/no.php/master/no.php
# Proxy requests like this: IN => BACKEND => OUT
# Your Node.js app must support running in this subdir (no links to /)
$backendUrl = 'http://localhost:81' . $_SERVER['REQUEST_URI'];

$headersIn = [];
foreach ($_SERVER as $key => $value) {
  if (!preg_match('/^HTTP_/', $key) || # only keep HTTP headers
    preg_match('/^HTTP_HOST/', $key) || # let curl set the actual host/proxy
    preg_match('/^HTTP_ORIGIN/', $key) ||
    preg_match('/^HTTP_CONTENT_LEN/', $key) # let curl set the actual content length
  ) {
[...]

Issue #145, opened 2016-12-08
Only allow users to edit and delete their own queries, unless admin


vweevers:
Would you be open to a PR? It can be an opt-in feature, if need be.

rickbergfalk:
Definitely! I'm not using this in a team setting at all, so I'm all for it being shaped by real-world use :)

vweevers:
@rickbergfalk your code hints at a future "viewer" role. What are/were your plans with that?

rickbergfalk:
It was going to be a role that could view and execute the queries as written by others, without the ability to modify the query prior to execution. I hadn't decided on whether the query editor UI would be presented to the user as is but locked down, or if It would be something different. 

Since the viewer role would only consume query results, I was thinking the query could run on open, kind of like how the grid-only/chart-only pages work. Maybe it would default to the chart view if a chart was configured. The actual SQL run could be viewable for reference, but it wouldn't be a primary focus when opening the query. 

I haven't really thought about this beyond that. I haven't had much time lately to flesh out my remaining sqlpad ideas. 

Also worth nothing - All my sqlpad has been as a solo user, so I'm not always entirely sure what's useful for a team and what isn't.

If this is something you are interested in feel free to make it however feel would be beneficial. I'm not invested in any particular vision of what this might be like in practice :)

lexadavid:
This feature would be amazing. My team started using SQLPad heavily and I worry that anybody can delete other people's queries. 
If only admin can do it, the problem is solved.

Thanks for great tool @rickbergfalk!

vweevers:
FIY I've implemented an ACL for SQLPad, with which one can grant or revoke a `principal` one or more `actions` to a `resource`. It looks like this:

```js
Rule.grant('role:admin', '*', '*')
Rule.grant('role:editor', 'query', ['create', 'read'])

// Can also add rules for specific users and resources
Rule.revoke('user:<userid>', 'query:<query id>', ['update']);
```

While not feature-complete, we've been using it in our team for about 1.5 months now, without problems. Hopefully I can find some time soon to push this upstream.

rickbergfalk:
Sorry for the delay in getting to this, but thanks for for the kind words @lexadavid! 

And @vweevers - the ACL would be amazing to add in. I'm all for pulling in your additions, especially given it is seeing real usage with your team.

davidmwilliams:
Hi @vweevers, were you able to push your mods?

rickbergfalk:
[...]

PR #459, opened 2019-09-01
Turn URLs in table columns into links

17dec:
This allows for easy linking to database entries in a different web interface.

FYI: We're running a [public SQLPad instance](https://vndb.org/d18) at VNDB.org with this patch applied. [Example of a common use case](https://query.vndb.org/query-table/bim26nb17M1UwjwR)

### DIFF
--- a/client/src/common/QueryResultDataTable.js
+++ b/client/src/common/QueryResultDataTable.js
@@ -18,6 +18,12 @@ const renderValue = (input, fieldMeta) => {
     return input.substring(0, 10);
   } else if (typeof input === 'object') {
     return JSON.stringify(input, null, 2);
+  } else if (typeof input === 'string' && input.match('^https?://')) {
+    return (
+      <a target="_blank" rel="noopener noreferrer" href={input}>
+        {input}
+      </a>
+    );
   } else {
     return input;
   }


rickbergfalk:
Thanks @17dec! For the contribution but also the info on how you're using it.

I skimmed the changes and poked around the links you sent, and will probably dive into things to understand the use case a bit more. If there's a configurable way to add the authentication mechanism you're using I'm open to accepting pull requests in that area. Various way of managing queries (private, shared, etc) too... actually if kept as simple as queries either being private or public that would be pretty easy to implement today.

SQLPad was started way back before I knew what I know now so there is a lot of debt and things to deal with 😅

17dec:
> If there's a configurable way to add the authentication mechanism you're using I'm open to accepting pull requests in that area.

Nah, the authentication changes I made are really specific to VNDB, they won't be useful elsewhere. However, what I'm doing is conceptually similar to the header authentication approach discussed in #301, and in fact that approach could work for me as well (but was more work for than just hacking in the VNDB auth directly). There's one caveat to consider with both my auth hacks and with header authentication: There is no sign-up and sign-in flow, and with this approach SQLPad's built-in user database and sessions are completely bypassed. Users are not stored and not queryable in the database and storing other data in the session is not possible (because there is no persistent "session"). This works fine with SQLPad right now, but I'm not sure if you intend this to be is a supported use case in the future. If you ever rely on the user list or proper sessions being available, then it's not a terribly complicated modification to make sure users are automatically added to the DB and session cookies are being created - but it's a little less efficient.

> if kept as simple as queries either being private or public that would be pretty easy to implement today

Nah, while some users might have a use for private queries, I personally don't mind that not being a feature. It's not very social, after all. :)

The current saving and sharing model actually works pretty well for us. The only (minor) change I made was to ensure that only the owner of a query could delete or edit said query, to prevent people from messing with each other's saved queries. This is basically a half-assed implementation of #145 - I hadn't reflected this change to the front-end code. If I ever find the time for a cleaner implementation I'd be happy to provide a PR.

PR #481, opened 2019-11-10
Adding MultiSubnetFailover option for SQL Server

jdrickerd:
This adds support for the MultiSubnetFailover connection option that allows connections to an AlwaysOn SQL Server cluster.


rickbergfalk:
Thanks @jdrickerd!

PR #484, opened 2019-11-26
Json result

gonace:
Added `.json` as a supported downloadable result, when a query is executed correctly one can now download the result as a JSON format.


rickbergfalk:
Thanks @gonace!

PR #489, opened 2019-12-03
Mysql ssl support

gonace:
* Added support for SSL under MySQL/MariaDB

Since there are no tests for other drivers and SSL support (e.g. PostgreSQL) I skipped this as well for MySQL/MariaDB. I've asked for help testing this out in the issue regarding this feature request.

### DIFF
--- a/server/drivers/mysql/index.js
+++ b/server/drivers/mysql/index.js
@@ -1,3 +1,4 @@
+const fs = require('fs');
 const mysql = require('mysql');
 const { formatSchemaQueryResults } = require('../utils');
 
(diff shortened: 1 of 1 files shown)


server/drivers/mysql/index.js
```diff
@@ -45,14 +46,26 @@ function runQuery(query, connection) {
...
+      key: fs.readFileSync(connection.mysqlKey),
+      cert: fs.readFileSync(connection.mysqlCert)
+    };
+    if (connection.mysqlCA) {
+      myConfig.ssl['ca'] = fs.readFileSync(connection.mysqlCA);
+    }
+  }
 
   return new Promise((resolve, reject) => {
+    const connection = mysql.createConnection(myConfig);
```
robofish: renaming this from `myConnection` to `connection` breaks the whole driver, as the name collides with the argument of the `runQuery` function which has the same name.
This leads to the later check `rows.length < connection.maxRows` *always* being falsy, due to `connection.maxRows` being `undefined`.
=> `rows` will always be empty and `incomplete` will always be true

PR #493, opened 2019-12-07
Remove whitespace at start of lines, add type hint

ltning:
When posting to anything else than Slack (haven't actually tested Slack), e.g. Mattermost, the leading whitespace makes output render funkily. Also, allow highlighting by specifying language. Minor cosmetic fix, but major usability improvement.


gonace:
Thanks @ltning! 👍

PR #494, opened 2019-12-13
Wait for admin password to get hashed before persist (new admin)

corneliusludmann:
Similar to #487 `await` is also needed a few lines below where default admin password is hashed for a new admin.


gonace:
Thanks @corneliusludmann ! 👍

PR #496, opened 2019-12-17
Set shell script interpreter to bash instead of sh

### LINKED ISSUE #495: Make build script more portable
To fix the issue with the missing `await` when using the admin password config, I used the [Gitpod Cloud IDE](https://gitpod.io/). I noticed that the build script does not run in Gitpod.

This is because `sh` is given as interpreter in the shebang line und double square brackets `[[ ... ]]` are used. In a POSIX `sh` double square brackets are not available. Most modern Linux desktop distribution does not provide a real `sh` but point to `bash` instead. Therefore using this script on a desktop works even though. Gitpod uses a docker container that has `sh` installed. Therefore, the script does not work correctly.

I would suggest to use `bash` as interpreter to make the build script more portable.

**By the way:** If you would like to provide other developers with an easier entry you could find a configuration for the Gitpod Cloud IDE with a running SQLPad in my [SQLPad Gitpod branch](https://github.com/corneliusludmann/sqlpad/tree/gitpod). Gitpod is free for up to 100 hours per month for open source projects. You can add it to your repository, if you like. _[Full disclosure: I work for TypeFox who makes Gitpod.]_

corneliusludmann:
On many systems sh points to bash. In that case the scripts work fine. However, double square brackets won't necessarily work when sh does not point to bash but to a POSIX sh. This commit changes the interpreter to bash. Alternatively, one can change the double square brackets to single square brackets to achieve portability.

Fixes #495.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 SQLPAD_CLIENT_DIR=$(pwd)/client
 SQLPAD_SERVER_DIR=$(pwd)/server
 SCRIPTS_DIR=$(pwd)/scripts
--- a/scripts/version.sh
+++ b/scripts/version.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 
 BRANCH=$(git rev-parse --abbrev-ref HEAD)
 if [[ "$BRANCH" != "master" ]]; then


rickbergfalk:
I'm not too experienced with bash or sh scripting... but would this change break for someone using zsh? (which is now the default in macOS, but I haven't bothered to switch over)

gonace:
This a good question, I'll have the possibility to test this at home in the next couple of days if @corneliusludmann is not able to verify or test this.

corneliusludmann:
As long as bash is installed it does not matter whether the default shell is zsh or any other shell. The first line (shebang) says exactly that, that _bash_ should be used for executing the script. I assume that MacOS will still have a bash installed. Since I don't use MacOS myself, I can't really say much about it.

rickbergfalk:
Just heads up - I went ahead and reverted this commit as it broke the automated docker builds.

The docker image uses Alpine linux, which doesn't use `bash` but instead uses `ash`?

```
Cloning into '.'...
Warning: Permanently added the RSA host key for IP address '192.30.253.113' to the list of known hosts.
Switched to a new branch 'v3.5.2'
Pulling cache layers for index.docker.io/sqlpad/sqlpad:3.5.1...
Done!
KernelVersion: 4.4.0-1060-aws
Components: [{u'Version': u'18.03.1-ee-3', u'Name': u'Engine', u'Details': {u'KernelVersion': u'4.4.0-1060-aws', u'Os': u'linux', u'BuildTime': u'2018-08-30T18:42:30.000000000+00:00', u'ApiVersion': u'1.37', u'MinAPIVersion': u'1.12', u'GitCommit': u'b9a5c95', u'Arch': u'amd64', u'Experimental': u'false', u'GoVersion': u'go1.10.2'}}]
Arch: amd64
BuildTime: 2018-08-30T18:42:30.000000000+00:00
ApiVersion: 1.37
Platform: {u'Name': u''}
Version: 18.03.1-ee-3
MinAPIVersion: 1.12
GitCommit: b9a5c95
Os: linux
GoVersion: go1.10.2
Starting build of index.docker.io/sqlpad/sqlpad:3.5.2...
Step 1/8 : FROM node:12.3.1-alpine
---> 91acf04599c4
Step 2/8 : ENV NODE_ENV production
---> Using cache
---> 37abc5f5c2dc
Step 3/8 : EXPOSE 3000
---> Using cache
---> 02564c4020ca
Step 4/8 : ENTRYPOINT ["/docker-entrypoint"]
---> Using cache
---> 1a788d8ab755
Step 5/8 : WORKDIR /sqlpad
---> Using cache
---> f44a9933cedd
Step 6/8 : COPY . .
---> bdec28c21aa1
Step 7/8 : RUN scripts/build.sh && npm cache clean --force && cp -r /sqlpad/server /usr/app && cp /sqlpad/docker-entrypoint / && chmod +x /docker-entrypoint && rm -rf /sqlpad
---> Running in 0fb5e2099e1c
env: can't execute 'bash': No such file or directory
Removing intermediate container 0fb5e2099e1c
The command '/bin/sh -c scripts/build.sh && npm cache clean --force && cp -r /sqlpad/server /usr/app && cp /sqlpad/docker-entrypoint / && chmod +x /docker-entrypoint && rm -rf /sqlpad' returned a non-zero code: 127
```

rickbergfalk:
Actually @corneliusludmann @deemoowoor is the existing shebang `#!/usr/bin/env sh` path an issue? Does `#! /bin/sh` work in your environment?

deemoowoor:
In this particular case `#!/usr/bin/env bash` should work best, I think. Using something like `#!/bin/bash` is usually fine, too, unless you're running the script on some very specific setups/systems where shell might be installed somewhere else than `/bin` (somewhat rarely, it might be).

deemoowoor:
Unless someone is willing to rewrite the script to POSIX shell syntax, changing it to use `#!/usr/bin/env bash` shebang should suffice, I think.

PR #510, opened 2020-01-04
Fix postgres multi-statement queries

koszti:
**Problem**
Running multi-statement queries in Postgres driver returns `Error: postgres.runQuery() must return rows array` error. To reproduce run the following query with a valid postgres connection:
```
SELECT 1;
SELECT 2;
```
This returns `Error: postgres.runQuery() must return rows array` error.

**Reason**
The `query` function in `pg.Client` that used to run multi-statements queries as a "workaround" returns array of `Result` objects but `runQuery` should return rows array. 

**Fix**
Transform array of Result objects to flat rows array.

### DIFF
--- a/server/drivers/postgres/index.js
+++ b/server/drivers/postgres/index.js
@@ -1,5 +1,6 @@
 const fs = require('fs');
 const pg = require('pg');
+const _ = require('lodash');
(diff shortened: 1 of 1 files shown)


rickbergfalk:
Thanks for catching this @koszti!

PR #511, opened 2020-01-04
Add Pre-query statements options to MySQL driver

koszti:
**Summary**

This PR adds optional **Pre-query statements** option to MySQL driver. This is one ore more SQL queries that run before the actual query but in the same session.  This is useful to enforce some session variables whenever it's required. Typical use-case is to avoid long running queries from SQLPad by setting session variables (ie. `SET max_statement_time = x`) before running the queries. 

Additionally added **Deny multiple statements queries** option to avoid to overwrite pre-query statements by running multi-statements queries from the UI. 

**Connection form with new items**
Screenshot 2020-01-04 at 14 30 36

### DIFF
--- a/client/src/common/TextArea.js
+++ b/client/src/common/TextArea.js
@@ -1,0 +1,20 @@
+import React from 'react';
+import styles from './TextArea.module.css';
+
+export default function TextArea({ children, error, className, ...rest }) {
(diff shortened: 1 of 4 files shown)


rickbergfalk:
Thanks @koszti!

PR #512, opened 2020-01-06
Add Query history

koszti:
## Description

Adding query history for audit-log purposes. Sometimes we need to know who ran specific queries at a certain time period or keep security and auditor teams happy with audit-logs. This PR creates `queryhistory.db` in NeDB and inserts new item with query details every time when a query executed. The UI provides a form to search certain items in the history.

To avoid large memory footprint introducing a new **queryHistoryRetentionTimeInDays** variable that defaults to 30 days. Every historical query after the retention time will be deleted and compacted automatically in NeDB. If set to 0 then queries not inserted into query history.

## Screenshots

1. New option in the menu:

Screenshot 2020-01-06 at 18 48 52

2. Filter options in query history window:

Screenshot 2020-01-06 at 18 47 21


server/models/queryHistory.js
```diff
@@ -0,0 +1,68 @@
+const db = require('../lib/db.js');
+const Joi = require('@hapi/joi');
+const config = require('../lib/config');
+
+const schema = Joi.object({
```
rickbergfalk: Can we also capture queryId if it exists? I think a nice enhancement following this is to have the history for a specific query available in case user wants to restore a previous version (though maybe that might get weird with partial query executions when highlighting a snippet of SQL?)

koszti: that makes sense, I will add it

koszti:
@rickbergfalk I have one more smaller pending PR on my list: Pre-query statements option for postgres. Similar to #511 but for pg.

Once this PR and the above one is completed (will send the PR later) we will have every required item to start using sqlpad on prod. I hope it will be useful for others as well.

rickbergfalk:
@koszti did you want to add query id to history in this PR or should I go ahead and merge as is? Ran this locally to try it out and I'd be okay either way.

koszti:
I'll update this PR and add the query id

koszti:
@rickbergfalk added `queryId` and `queryName` to query history by 6df0666aa79d944b08f6f67f1283668d6173c0c5. Added some extra tests as well and now this PR is completed.

server/models/queryHistory.js
koszti: added by 6df0666

rickbergfalk:
Thanks @koszti!

PR #513, opened 2020-01-09
Update "odbc" version to latest which builds with Node 12.

rlwhitcomb:


rlwhitcomb:
Ugh. Travis build uses Node 8, probably need to do "npm install" anyway to update package-lock as well.

PR #514, opened 2020-01-09
Update odbc version to 2.2.2 so it can be built inside docker container

rlwhitcomb:
Docker image is using node 12, while odbc version 1.4.1 requires node 8. Update Travis settings to build with node 12 also.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 services:
   - docker
 node_js:
-  - '8'
+  - '12'
 before_install:
   - sudo service mysql stop
   - sudo service postgresql stop
--- a/server/drivers/unixodbc/index.js
+++ b/server/drivers/unixodbc/index.js
@@ -118,7 +118,7 @@ const fields = [
     key: 'schema_sql',
     formType: 'TEXT',
     label:
-      'Database sql to lookup schema (optional, if ommited default to checking INFORMATION_SCHEMA)'
+      'Database SQL to lookup schema (optional, if omitted default to checking INFORMATION_SCHEMA)'
   },
   {
     key: 'username',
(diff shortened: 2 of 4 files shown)


rickbergfalk:
Thanks @rlwhitcomb - the major version bump typically indicates breaking API changes, though that could be due to the node 12 requirement too. Do you happen to know whether the ODBC driver is still functional with SQLPad during runtime? 

If unsure I can take a look and verify this weekend.

rlwhitcomb:
@rickbergfalk I will try to verify ODBC today, still working to get this going inside one of our clusters.

rickbergfalk:
@rlwhitcomb it looks like odbc version 2.x has significant API changes. I've updated the SQLPad driver implementation for this to work in https://github.com/rickbergfalk/sqlpad/tree/pr/514. 

Something I'm wrestling with is what to do regarding minimum Node.js version supported relating to odbc.

Node 10 is a LTS branch and supported until April 2021 (https://nodejs.org/en/about/releases/). Having Node 12 be the minimum supported version for SQLPad (for odbc users at least) would also maybe require a major version bump in SQLPad to properly follow semver?

Options we have moving forward:

* Leave odbc lib at v1, change container to use Node 10, and limit/require odbc users to use Node 10 when running SQLPad
* Update odbc to v2, change container to use Node 12, and set Node 12 as minimum supported version for SQLPad project (and maybe bump SQLPad's major version?)

Any thoughts on how to proceed? @koszti @gonace ?

gonace:
@rickbergfalk I'd vote for moving to Node 12 and bumping a major version. 👍

rickbergfalk:
@rlwhitcomb I pulled this PR into #516 and updated the odbc driver implementation so it is compatible.

@gonace I'll cut new release and bump major version of SQLPad to indicate the node 12 requirement

PR #520, opened 2020-01-27
Adding MSSQL ReadOnly option

jdrickerd:
Adding the ReadOnly Application Intent option for connecting to AlwaysOn clusters.


gonace:
@jdrickerd It would be nice if this could be tested, what do you think @rickbergfalk ?

jdrickerd:
It would be nice to be tested. I wasn't sure how to test it since it is for connecting to an availability group. I'm open for ideas though.

jdrickerd:
When I added the other driver option for MultiSubnetFailover, we didn't add tests for that either for the same reason. These are just mapping the options available in the driver.

rickbergfalk:
Sorry for the delay in getting to this! Somehow missed it first time through.

For driver specific config enhancements I've been leaving testing at the core driver config only. My assumption is if the core configuration still works, the additional fields are okay to add.

I'd like to be able to test something like this but I just don't know how realistic it is to be able to do with docker...

Issue #532, opened 2020-02-18
Reference user info in connection configuration


rickbergfalk:
I'd like to enhance connection configuration to be able to reference user data, dynamically applying user values to the connection when running a query. 

An example of this might be associating a database username with a SQLPad user, and using that value when running queries for that connection. This would allow a single connection configured in SQLPad, but in practice the connection would be unique to that user. 

To support custom user info, a generic `data` json field will be added to SQLPad's user model, and the UI extended to add arbitrary attributes to the user model that are not "owned" by SQLPad.

A mustache/handlebars template library will assist with value. For example, setting a connection's `username` field to a custom `dbuser` value might look like `{{{user.data.dbuser}}}`.

PR #544, opened 2020-02-20
Add connection template support

rickbergfalk:
Adds the ability to reference user info in connection configuration using a mustache/handlebars like syntax. 

For any text (string) connection field, user info may be referenced using `{{user.fieldName}}` syntax. 

This also adds additional user fields, `name` and `data`. Name will be used eventually for a the preferred human friendly user name. `data` is to be used to store whatever end users would like to store, and is not something SQLPad will be caring about.

Example: Let's say each SQLPad user has a specific database username they are to use when connecting to a database. You may specify this value in the SQLPad `user.data` field as `user.data.dbuser` (how that is to happen TBD), then reference that value when defining a connection, setting username to `{{user.data.dbuser}}`. When queries are run against a database using that connection, `{{user.data.dbuser}}` gets swapped out with the value associated with the user running the query.

### DIFF
--- a/server/drivers/index.js
+++ b/server/drivers/index.js
@@ -1,4 +1,5 @@
 const uuid = require('uuid');
+const _ = require('lodash');
 const config = require('../lib/config');
 const utils = require('./utils');
 const getMeta = require('../lib/getMeta');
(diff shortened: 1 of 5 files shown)


rickbergfalk:
This implements #532, but does not add any UI associated with this feature.

PR #546, opened 2020-02-21
Added support for html in data rows

gonace:


gonace:
Will reopen when the feature is done

PR #549, opened 2020-02-21
Add custom application header

rickbergfalk:
Adds a config option to specify a mustache template to be used for a custom application header rendered above the editor toolbar. The template can reference `name`, `email`, and `data` fields of the user, for cases where the logged in user info is desired. 

The template path may be specified using the key `appHeaderTemplate` and environment variable `SQLPAD_APP_HEADER_TEMPLATE`. 

## Example

The following template file is saved under `server/test/fixtures/app-header-template.html` and provided via ini setting `appHeaderTemplate = "./test/fixtures/app-header-template.html"`.

```html
<style>
  .custom-header-root {
    background-color: rgba(0, 0, 0, 0.82);
    border-bottom: 1px solid rgb(204, 204, 204);
    color: #fff;
    display: flex;
    height: 40px;
    justify-content: space-between;
    padding: 8px;
    text-shadow: 0 -1px 0 rgba(0, 0, 0, 0.12);
  }

  .custom-header-link {
    color: #fff;
  }
</style>
<div class="custom-header-root">
  <div><a href="/" class="custom-header-link">&lsaquo; Link to somewhere</a></div>
  <div>My Custom SQLPad</div>
  <div>{{user.name}}</div>
</div>
```

It results in the following:

Screen Shot 2020-02-21 at 3 29 39 PM

If no template is specified, no application header is rendered.

### DIFF
--- a/client/src/queryEditor/CustomAppHeader.js
+++ b/client/src/queryEditor/CustomAppHeader.js
@@ -1,0 +1,38 @@
+import React, { useEffect, useState } from 'react';
+import { connect } from 'unistore/react';
+import fetchHtml from '../utilities/fetchHtml';
+
+function mapStateToProps(state) {
+  return {
+    config: state.config
+  };
+}
+
+const ConnectedCustomAppHeader = connect(mapStateToProps)(
+  React.memo(CustomAppHeader)
+);
+
+function CustomAppHeader({ config }) {
+  const [headerHtml, setHeaderHtml] = useState({ __html: '' });
+
+  useEffect(() => {
+    if (config.customAppTemplateConfigured) {
+      let subscribed = true;
+      fetchHtml('/custom-app-header').then(result => {
+        if (subscribed) {
+          setHeaderHtml({ __html: result });
+        }
+      });
+      return () => {
+        subscribed = false;
+      };
+    }
+  }, [config]);
+
+  if (headerHtml && headerHtml.__html) {
+    return <div dangerouslySetInnerHTML={headerHtml} />;
+  }
+  return null;
+}
+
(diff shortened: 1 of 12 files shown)


rickbergfalk:
@gonace @koszti have any thoughts on this approach? 

Something I'm torn about is that a typical use case of this header would be to put the user name/email up in the right-hand side, as applications tend to do. This is fine, but the menu button in the toolbar that holds application things seems to be in the wrong place once the header is added. It almost seems like that menu button should be moved into the application header to fit the information hierarchy better? 

Same goes for the queries & new query buttons.

SQLPad v2 had a 2 level structure like this, but I removed the application header in v3 to maximize vertical space. I also had removed the other "pages" in the application, moving things to drawers and dialogs, so the application header didn't seem as important.

So I guess what I'm thinking about is... Instead of a custom header where anything goes, do we formally bring it back with some limited customization?

The use case for this work is to aid an org in embedding SQLPad in their environment, providing a custom title/logo, show logged in user, and provide a link back to a main portal for the user.

I think I'm going to take the weekend and think about this a bit before merging. I'd also be fine merging this, then seeing how much traction this gets, and bringing a formal application header if that is of interest?

gonace:
Hi @rickbergfalk 
Would this not be an opportunity to redo the navigation a bit? At least I feel that it would be nice to figure out a way so that the upper "navigation" fits more into the application. 

A rough example of what I'm trying to say:
nav

rickbergfalk:
Yeah I'm kind of leaning towards that direction as well. I like the direction of the app bar in the mock!  

I'm going to hold off on merging this.

PR #550, opened 2020-02-24
Add a build pipeline

rgarrigue:
This PR add a Github Action build pipeline to SQLPad. 

Every tag on the repository will trigger the usual build, tar the result and add it to Github Release next to the source code, like this https://github.com/rgarrigue/sqlpad/releases/tag/v4.0.2

This will allow users to install SQLPad on a basic web server downloading / untaring / adding a config file without installing all the build dependencies.

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
 SQLPAD_CLIENT_DIR=$(pwd)/client
 SQLPAD_SERVER_DIR=$(pwd)/server
 SCRIPTS_DIR=$(pwd)/scripts
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,12 +1,11 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
 SQLPAD_CLIENT_DIR=$(pwd)/client
 SQLPAD_SERVER_DIR=$(pwd)/server
 SCRIPTS_DIR=$(pwd)/scripts
 
-if [[ ! -d $SCRIPTS_DIR ]] || \
-   [[ ! -d $SQLPAD_CLIENT_DIR ]] || \
-   [[ ! -d $SQLPAD_SERVER_DIR ]]
-then
+if [ ! -d $SCRIPTS_DIR ] ||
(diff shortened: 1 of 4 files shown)


scripts/build.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env sh
+#!/usr/bin/env bash
```
rickbergfalk: Does this need to change to bash? 

This was previously changed to bash but had to be reverted as the official docker image runs alpine and does not have bash installed.

rgarrigue: Then you shouldn't use [[ ]], sh doesn't know this syntax. I reverted the shebang in the following commit and replaced the [[ ]] by simple [ ], which works too

rgarrigue:
@rickbergfalk would be glad to have a tag following merging this one btw ! Maybe adding a quick note in the readme about it would be good enough if you don't have anything else worth tagging

scripts/build.sh
rickbergfalk: Thanks! My sh/bash skills aren't great but the scripts have seemed to work in my environments so far. Thanks for fixing this too :)

rickbergfalk:
@rgarrigue will get this tagged and try this out. There's quite a few things to mention so its good timing.

rickbergfalk:
Haven't looked into it too much but looks like there was an issue uploading some of the assets? Looks like source code for server and client files are there, but the client build file is not.

```
Run actions/upload-release-asset@v1
  with:
    upload_url: https://uploads.github.com/repos/rickbergfalk/sqlpad/releases/23977940/assets{?name,label}
    asset_path: ./sqlpad.tar.gz
    asset_name: sqlpad.tar.gz
    asset_content_type: application/gzip
  env:
    GITHUB_TOKEN: ***
##[error]Error saving asset
```

rickbergfalk:
Alright deleted the release and re-ran and things seemed to work out. Thanks again for this @rgarrigue!

rgarrigue:
Yes, I've been using Github Actions for a couple of week for similar usage, sometime it fails, not sure why. You can trigger it again in the Actions tab and/or remove the tag and create it again, doesn't change much imo.

Anyway, you're welcome, thanks for SQLPad :)

PR #553, opened 2020-02-25
Test infrastructure improvements

rickbergfalk:
A bunch of testing infrastructure improvements to take advantage of the recent config, app, and models refactoring. 

* Restructures test utils to be a class, containing references to its own database and application objects. Each testUtils instance should technically be its own isolated environment!!!
* Adds a `dbInMemory` config item to create nedb databases in memory. `dbPath` is still required for cache and session support at this time.
* Adds a `wait` setting to mock driver to reduce faked wait time when running a mock query
* `fromEnv` and `fromFile` no longer have a default value provided. It is expected something is always provided like `fromCli`

With these changes each test case should be able to stand on its own, and not rely on a "global" test config. The general test cases will go something like this:

```js
const assert = require('assert');
const TestUtils = require('../utils');

describe('api/tags', function() {
  // This collection has its own private nedb automatically created
  const utils = new TestUtils();

  // before anything is done test utils must be initialized
  // the parameter adds in users (effectively `resetWithUsers` but with a fresh db)
  before(function() {
    return utils.init(true);
  });

  // assert, make API calls, whatever
  it('Performs some test', async function() {
    const res = await utils.post();
  });
});

```

### DIFF
--- a/server/drivers/mock/index.js
+++ b/server/drivers/mock/index.js
@@ -89,7 +89,8 @@ function cartesianify(rows, field) {
  * @param {object} connection
  */
 async function runQuery(query, connection) {
-  await wait(2000);
+  const ms = connection.hasOwnProperty('wait') ? connection.wait : 2000;
+  await wait(ms);
 
   // Connection here doesn't actually matter.
   // Someday this mock could get fancy and change output based on some connection value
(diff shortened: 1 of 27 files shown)


rickbergfalk:
Haven't looked into it too much but looks like there was an issue uploading some of the assets? Looks like source code for server and client files are there, but the client build file is not.

```
Run actions/upload-release-asset@v1
  with:
    upload_url: https://uploads.github.com/repos/rickbergfalk/sqlpad/releases/23977940/assets{?name,label}
    asset_path: ./sqlpad.tar.gz
    asset_name: sqlpad.tar.gz
    asset_content_type: application/gzip
  env:
    GITHUB_TOKEN: ***
##[error]Error saving asset
```

PR #554, opened 2020-02-26
Add Query ACL (and migrations framework)

rickbergfalk:
Adds initial query ACL implementation, as well as related infrastructure for doing so via SQLite, sequelize ORM, and umzug migration tooling.

## Query ACL

By default `admin` users have full access to all queries, and query creator's have full access to their own query. For everyone else, access is determined by query ACL entries.

The query ACL is a listing of records containing `queryId`, `userId`, and a `write` flag. If a row exists for a user, it provides the user access to read the query. Execution permission is dependent on connection access. The user can only save the query if the `write` flag is set to true.

Access permitted via ACL does not permit the query to be deleted. Only query creator or Admin may delete the query.

A special `__EVERYONE__` userId entry may be made for a query to set access to everyone.

**Pre-existing queries to this change will be migrated to give `__EVERYONE__` write access to the queries. However, all future queries created will be private by default.**

The query ACL is the first model to be stored in SQLite.

For this initial implementation, the UI is only exposing a single share toggle button. Either the query is private, or it is shared to everyone with write permissions. A more advanced UI will follow at some point.

The query list has verbiage changed, from showing teams queries, to queries "Shared with me".

Screen Shot 2020-02-26 at 1 09 21 PM
Screen Shot 2020-02-26 at 1 09 33 PM
Screen Shot 2020-02-26 at 1 08 50 PM
Screen Shot 2020-02-26 at 1 08 59 PM

## Migrations

As of this change, migrations are run during server start. Migrations are run in alphanumeric order. Once a release it tagged, migrations are not expected to change unless absolutely necessary. Fixing migrations that fail to run is fine, but fixing migrations the yield an incorrect result is not. Fixing bad data must be done with a follow-up migration.

As of this PR migrations may alter data in nedb or SQLite, but the record of schema version is kept in SQLite. 

Migration versions are numbered based on current major version of SQLPad, along with some additional change version number.  These change version numbers are arbitrarily spaced and padded in case additional migrations need to be written to precede existing migrations.


rickbergfalk:
@gonace @koszti Either of you want to take a look at this before I merge?

I'm going to step away from it for a bit and self-review tomorrow, and merge if there are no concerns.

This was initially going to be just for the migration tooling, but it was hard to know what that looked like without doing the next work using sequelize/sqlite.

PR #561, opened 2020-02-29
UnixODBC: Multiple statement support

rickbergfalk:
Adds multiple statement/batch query support to the unix odbc driver implementation. This is accomplished by splitting the batch SQL into statements using the query separator `;`. Each statement is executed sequentially using a single connection to the database.

The query result from the last statement that had rows returned is sent to the client. All statements are run, unless there is an error, in which case running statements in the batch stops and the underlying connection is closed.

This adds a `suppressedResultSet ` indicator to drivers.runQuery for cases like this, as long as a corresponding indicator in the UI similar to incomplete indicator.

Screen Shot 2020-02-28 at 8 40 47 PM

Long term the plan is to show multiple result sets in SQLPad. This is just a step towards getting there. Supporting large query batches may require redesigns to the query REST API (or at the very least setting smaller row-limits per query).

### DIFF
--- a/client/src/QueryChartOnly.js
+++ b/client/src/QueryChartOnly.js
@@ -2,6 +2,7 @@ import PropTypes from 'prop-types';
 import React, { useEffect, useState } from 'react';
 import ExportButton from './common/ExportButton.js';
 import IncompleteDataNotification from './common/IncompleteDataNotification';
+import SuppressedSetNotification from './common/SuppressedSetNotification';
 import SqlpadTauChart from './common/SqlpadTauChart.js';
 import { exportPng } from './common/tauChartRef';
 import fetchJson from './utilities/fetch-json.js';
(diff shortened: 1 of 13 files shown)


rickbergfalk:
I've noticed in testing with sqlite that `;` in comments are detected as the query having "multiple statements" and are not allowed. 

For example this query is not allowed

```sql
-- This is a comment ;
SELECT 'hello' as world; 
```

But this one is allowed

```sql
-- This is a comment
SELECT 'hello' as world;
```

SQLPad can detect the `;` in comment blocks in strip it out prior to running the query, but should it? Is this a node-odbc implementation issue or the underlying driver being referenced (in this case SQLite3 ODBC Driver in Windows).

@clach04 I'll keep digging but you may find this of interest

clach04:
@rickbergfalk hmmmm, my  educated guess is that this is specific to the SQLite3/ODBC combo (unrelated to node-ODBC) and that its's not SQLpad's job to do that. If you did decide you wanted to do that in SQLpad I would recommend it be via a flag that the user has control over (e.g. for sqlite specifically). For example, Oracle uses comments to indicate optimizer "hints".

Have you got access to a Vector server (or Avalanche) to test with the Actian ODBC driver? That should give you confirmation (or not ;-)) of my claims above. PostgreSQL, MySQL, etc. could also be used to test.

We should be able to get you up and running with a local Vector server if you are interested - its super fast so if you have any OLAP testing to do you may find it useful. Email me directly if interested.

PR #568, opened 2020-03-06
Add Application header bar

rickbergfalk:
Adds an App header similar to what was mocked in #549 

Not set on the placement of everything exactly, but I'll probably make this configurable so certain buttons could be hidden or rearranged.

@gonace What do you think? I'm trying to think of what to have the blue SQLPad button do... should it be the new query button or open queries or not be clickable or something else entirely? I suppose it could open the about modal too. 

SQLPad header


rickbergfalk:
I've tweaked the design a bit from initial look from when I opened the PR. 

* Wasn't sure what to make the logo do when clicked, so I made it non-clickable
* I've dialed back the color for the "logo" section. Using primary blue kept making me think it was clickable and felt like it was taking attention from the run button
* The user "button" when clicked opened a drop down to sign out. It didn't feel right having that with the only option, so I moved it and the "about" button into the generic menu button on far right.

I think this is a good step towards having a traditional app header that can grow and have more traditional use. Eventually clicking the user can show sign out + a manage profile option and whatever else.

SQLPad header2

PR #572, opened 2020-03-10
Add BigQuery driver

murphyke:
@rickbergfalk Here's what I have so far for the BigQuery driver.

There is a README describing how to test, etc.

There are two things that I'd like to call out:

- Since the BigQuery API allows you to set the timeout on the query, I thought it would be good to use the SQLPad query timeout setting for that purpose. However, I didn't see a formal way for database drivers to access the SQLPad config settings, so in the short term I just copied how it's done elsewhere in the code base.

- Also related to query timeouts: several weeks ago, my current approach of throwing an exception for the timeout worked well: as soon as the timeout occurred, my custom error message would be displayed in the SQLPad results area. However, now there is a multi-second delay following the timeout, and the message that appears in the results area is generic: "Server responded not ok".

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,15 +22,67 @@
         "chalk": "^2.0.0",
         "esutils": "^2.0.2",
         "js-tokens": "^4.0.0"
+      },
+      "dependencies": {
+        "ansi-styles": {
+          "version": "3.2.1",
+          "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-3.2.1.tgz",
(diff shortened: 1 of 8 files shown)


rickbergfalk:
Thanks @murphyke! This looks great.

What you've done for the config works. I'll have to think about how to pass config to driver implementations.

Your code looks correct, so you should see the error in the UI if you are throwing. The message you are getting is a generic message from the front-end code when the response is something other than 200. Could you check the response and see what it is sending back? 

I'm wondering if the BigQuery timeout is happening after the SQLPad timeout, since SQLPad's timer probably starts sooner? You could try decreasing the timeout that is sent to BigQuery by a few seconds to see if that helps.

murphyke:
> You could try decreasing the timeout that is sent to BigQuery by a few seconds to see if that helps.

Duh, thanks, that would do it. I am now dropping a second off the timeout, and that behavior is documented.

rickbergfalk:
Glad to hear that's all it was! This is excellent. Thanks again @murphyke

Issue #402, opened 2019-02-04
Create option to disable authentication entirely


akatch:
Wait, hear me out!

In March of 2019, Google+ Oauth will be deprecated, and thus will no longer work with SQLpad, so I've been looking at different ways to handle SQLpad authentication. One of the things my institution already uses for authentication is oauth2_proxy (with regular Google Oauth2, which is apparently different from Google+ Oauth) behind an Nginx `auth_request`. The oauth2_proxy / auth_request part of this seems to work just fine with SQLpad, however, it doesn't look like there is a way to disable the built-in authentication, so once a user authenticates with oauth2_proxy, they still hit the Google+ oauth or the userpass auth page. Is there something I'm missing, or is this a feature for which PRs might be accepted?

rickbergfalk:
Hey @akatch 

I don't actually use the Oauth auth in SQLPad, but I think it should support Oauth2 directly, so I think SQLPad's google-based auth should still work after March 2019...

I'm not 100% certain on how the proxy works, but what I think might be necessary is SQLPad reading some headers and tokens from the proxied request, and then trust those values and map them to a user (or something like that).

Disabling authentication entirely is something I'd be up for considering though... but I'm wondering how the users concept should be handled in SQLPad.

If authentication was disabled, would you expect there not to be a SQLPad user to be logged in at all and it'd be some shared identity? Or would you expect there to be some mapping involved... like an email address or identifier being read from the proxied request?

rickbergfalk:
Also out of curiosity, what does your SQLPad use case look like? There is an interesting discussion/request in #314 to configure connections via environment variables and disable authentication, using SQLPad for a shared dev tool of sorts.

akatch:
Hey @rickbergfalk, thanks a bunch for your response.

Even if SQLpad does support Oauth2 directly, I would still ideally like to get our instance behind our centralized oauth2_proxy setup, just for consistency across our environment. To that end, a set of hardcoded credentials like what's described in #314 would probably suit us fine. I honestly hadn't given any thought to how the concept of users in SQLpad would be handled if authentication were disabled, but after some consideration, I have a couple of ideas.

First, in a "development" environment, the concept of a separate admin user might not exist at all - access to the SQLpad instance would grant admin access. I can think of arguments both for and against this.

[...]

PR #576, opened 2020-03-12
Disable Authentication

hessam61:
@rickbergfalk This PR addresses the conversation in issue #402 by adding a new configuration `disableAuth` and creating a `noauth` user object through a middleware as you suggested in order to disable authentication entirely. It also removes all user-specific elements in UI.


rickbergfalk:
Thanks @hessam61! 

Would it be possible to revert the changes to the package-lock.json file by chance? This sometimes changes if `npm install` was run instead of `npm ci`. 

PR looks great otherwise 👍

hessam61:
Absolutely @rickbergfalk, and thanks for your review! I reverted the package-lock.json back to what's on the `master` branch right now.

rickbergfalk:
Excellent! Thanks again for your work on this!

PR #589, opened 2020-03-23
Test cleanup - improve config and local auth

rickbergfalk:
I'm working on implementing something similar to https://grafana.com/docs/grafana/latest/auth/auth-proxy/, and realized that the current tests don't work great for testing different authentication methods.

This PR has a few improvements for config/testing related things:
* `make<SomeAuth>` routes now instantiate router inside the function. It was otherwise being reused, causing issues when testing with different configs
* Consolidate config unit tests, leaving single argv param for testUtil
* Consolidate local auth tests to not rely on pre-defined user vars in test utils (now uses the user created during signup to signin).


rickbergfalk:
fyi @koszti I don't think this conflicts with anything you are working on but wanted to make you aware.

PR #588, opened 2020-03-23
Add service tokens

koszti:
Add option to generate Service Tokens to programatically access API endpoints using secure [JWT](https://jwt.io/) based authentication. Service Tokens meant to use for service-to-service communication, integration and automation of different services. Service tokens don't required pre-authentication to SqlPad, the credentials and permissions embedded in the token and it can communicate with the SqlPad API using single HTTP request.

Screenshot 2020-03-23 at 00 30 38

The service tokens are [JSON Web Tokens](https://jwt.io/) that can be generated in the UI. We can specify, **role** and **expiry** date for the token to keep every endpoint "role based" and secure even if it's accessed by services.

PR #590, opened 2020-03-23
Add proxy authentication support

rickbergfalk:
Adds auth proxy authentication support, inspired by https://grafana.com/docs/grafana/latest/auth/auth-proxy/

SQLPad's implementation is a bit different but the general idea is the same.

Config keys added as INI format:

```ini
; Enable auth proxy authentication
authProxyEnabled = true
; Auto create user record if it does not exist
authProxyAutoSignUp = true
; default role to use if not provided by header
authProxyDefaultRole = editor
; header mappings space-delimited. 
; convention is <user-field-to-map-to>:<header-name-to-use-for-value>
authProxyHeaders = "id:X-WEBAUTH-ID email:X-WEBAUTH-EMAIL name:X-WEBAUTH-NAME role:X-WEBAUTH-ROLE data.customField:X-WEBAUTH-CUSTOM-FIELD"
```

The extent of how much of this to use is flexible. At the very least only an email header is necessary. It'll be used for user lookups, as email is treated as a natural id. 

If auto sign up is turned on, then a default role needs to be provided, either via config key or via a header mapping.

Additional user fields may be specified, including custom data fields that will populate the `user.data` object. (This is useful for dynamic connections, something mostly implemented but not documented). 

This PR also updates the testing infrastructure to use the proxy auth approach. Doing so cuts down the test run times, as username/password encryption is no longer necessary.

(This PR is finished except for documentation)


server/middleware/must-be-authenticated.js
```diff
@@ -3,5 +3,7 @@ module.exports = function mustBeAuthenticated(req, res, next) {
   if (req.isAuthenticated()) {
     return next();
   }
+  // TODO send 401 instead
+  // Sending redirect is old behavior, and signin may not even be available based on auth setup
   res.redirect(req.config.get('baseUrl') + '/signin');
```
koszti: +1 to send 401 instead. I was also coming across with this when working on PR #588

rickbergfalk: 😅 yeah I think it'll be a separate PR. This tripped me up too during testing.

There's actually a lot of API stuff I'd like to fix. I'm kind of thinking about making a list and making it a v5 thing?

koszti: v5 is fine, we can also document the API and every endpoint. I know it's a good amount of work :)

PR #593, opened 2020-03-28
Separate auth strategy implementations from routes/middlewares

rickbergfalk:
This separates auth strategy implementations from routes/middlewares that use them, in hopes of providing clarity. This moves code around, there should be no logic changes in any of this.

I'm anticipating that various routes and middlewares will start to cross paths (For example, perhaps various strategy authenticate calls move to must be authenticated middleware).

I'm also planning to allow auth-proxy to use the signin route if implementations would like to lean on using the cookie-based session.

@koszti I noticed the JWT implementation was adding the strategy on each request. I'm guessing passport handles this just fine, but to be consistent with other strategy implementations I moved it  out from the middleware.


rickbergfalk:
I've also removed dynamically adding auth routes if a given auth is configured. The strategy won't exist  if it is not enabled, and additional handling can happen inside the route handler if necessary.

rickbergfalk:
While working on adding session support to auth-proxy stuff I realized it'll be more difficult to add than I thought 😓 

Ultimately what I'd like to get to is `/signin` authenticates using auth-proxy (like it would for other routes passively), but authenticates and starts a session. I don't know if that's possible though, because the auth proxy authentication happens passively for every route if headers are provided (so the `/signin` route would be authenticated twice. Even if that works, it feels gross to me.

I then thought I'd move the middleware authentications (like HTTP basic, auth-proxy) to the must-be-authenticated middleware, which I think will work too, but I wasn't sure if there are any routes that don't *require* authentication, but might prefer it if available.

This brought me to the table/chart link requires no auth setting, and I realized it was broken from a variety of feature additions (which is understandable as the use case wasn't tested)

So I've added some tests and a fix for those cases, and some comments.

I'm leaning towards deprecating the table/chart not requiring auth. It makes it hard to apply some broad security checks, and is becoming kind of a mess now that we have connection access and query ACL restrictions that need to key of a "user" of some kind.

Instead I think it makes sense that people needing no-auth to those routes use one of the enhanced methods of access in v5 -- like auth proxy, or JWT token or something.

PR #620, opened 2020-04-25
Sequelize/SQLite Implementation

rickbergfalk:
This is still broken. Just checking on progress.


rickbergfalk:
96 FILES CHANGED!?!?! I guess I won't feel bad when prettier changes them all after this.

rickbergfalk:
As of this most everything works! There's just some minor quirks with error states that need to be cleaned up, but this is a v5 thing, not necessarily tied to the sqlite/sequelize stuff

PR #622, opened 2020-04-27
Merge v5 to master

rickbergfalk:
v5 work is ready to merge to master. 

This contains mostly maintenance and refactoring, along with the removal of some deprecated configuration items.

At a high level:

* Improvement to REST APIs to use proper HTTP response codes and remove reponse "enveloping"
* Migrates data off nedb to SQLite
* Uses Sequelize as an ORM/query interfact
* Removes legacy /queries table/chart redirect 
* Removes `tableChartLinksRequireAuth` config
* Encrypts all of driver-specific connection values using `cryptr` (to replace deprecated approach)
* Removes bright/bold magenta/blue backgrounds and shadows, in favor of more subtle styles


rickbergfalk:
@bruth After merging this, running a build from master (docker `latest` tag) will migrate nedb data to SQLite. 

I recommend making a backup of dbPath dir, just in case there are any bumps with the migration, but I think its ready to try out if you're interested.

bruth:
Wow that was quick! Looking forward to trying it this week, thanks!

bruth:
@rickbergfalk Sorry for the delay. It appears to have been seamless on a test database. One question I have is after a fresh startup, I still see the existing `*.db` files and `sessions/` directory. I do see the `sqlpad.sqlite` file as well which I presume is the new single database?

rickbergfalk:
No worries -- yeah the `sqlpad.sqlite` file now the primary source of data. The `cache` directory still holds query result files, though that will be moving to the sqlite file soon as well, with the results being stored once as a compressed blob.

The `nedb` store is configured to autocreate the database files, so they are still created on start up. 

Something I need to think further about is migration management. Currently migrations run on startup, and are a part of the sqlpad process, so nedb is necessary for as long as migration from the old format is to be supported. 

I'm starting to think it might be best to have a separate migration process?

bruth:
Thanks. The "on-startup" migration is definitely the most seamless, but you are right in that the `nedb` dependency will need to linger. Approaches I have seen in past projects is to cap the version at which the migration will occur. Subsequent versions will remove this automated migration path, so it would require a step version upgrade before going to the latest. For example, `v5` lands the migration.. so all versions up to `v6` have the migration path. However `v6` would not include this code. So if a user is migrating from `v4` and wants to migrate, they would first need to run a `v5` instance to perform the migration prior to going to `v6`.

The other option would be a separate standalone (and idempotent) script that does the migration. This can be executed by the operator prior to starting up the main process. They can choose to do it as a one-off or as a prior step to the main process. Even with a `v6`, the script could still work if migrating from a `v4`.

rickbergfalk:
Do you have a preference on standalone script vs on-startup at all?

bruth:
Not really. The on-startup is more seamless, but knowing that a standalone script can be once and done external to the main process is mildly comforting as well. It largely depends on how isolate the migration is on startup so that it really doesn't botch anything in a future release accidentally or "re-migrates" old DB files accidentally.

PR #635, opened 2020-05-19
Issue 634 - BigQuery column descriptions

murphyke:
@rickbergfalk So this was simple except that some of our columns require long descriptions that can't be seen fully even when the sidebar is maximized.

You will see that I had to tweak your existing Tooltip component in order to get wrapped text. The underlying Reach UI Tooltip doesn't behave well with long text, but at least it allows style to be passed to it. (The Reach UI Tooltip sets `white-space` to `nowrap` and displays long tooltips truncated on one line ... oh, well.)

Here are a couple of screenshots:

**Basic column description display**:

bigquery-col-desc

** Hovering over the second column description**:

bigquery-col-desc-tooltip

### DIFF
--- a/client/src/common/Tooltip.js
+++ b/client/src/common/Tooltip.js
@@ -2,6 +2,6 @@ import React from 'react';
 import ReachTooltip from '@reach/tooltip';
 import '@reach/tooltip/styles.css';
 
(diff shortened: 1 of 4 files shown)


rickbergfalk:
Looks great - Thanks @murphyke!

PR #639, opened 2020-05-28
remove duplication

eladeyal-intel:


rickbergfalk:
Thanks @eladeyal-intel!

PR #644, opened 2020-05-31
remove duplicate

eladeyal-intel:
this looks duplicate to me


rickbergfalk:
Yep - the initial docs were generated from this file so it makes sense it was here too.

Issue #645, opened 2020-05-31
non-admin password-less login


eladeyal-intel:
My usecase for sqlpad is "minimal interaction" - click a URL and write sql.

When I set `DISABLE_AUTH=true`, everyone can login (which is good) but everyone get Admin access (which is not good). I would prefer an option to auto-login as Editor. (Connections are configured via env vars).

eladeyal-intel:
I tried to implement that in #646.

I guess it would be slightly better if I could have a "secret URL" - i.e. a URL with a token which will let users connect as guests (and others still can connect as admin).

rickbergfalk:
I've thought about this too and I think its reasonable to have the disable auth role configurable. 

Guest access is something I think would be good to add as well but I want to try and consider all the different use cases. Something I'd like to implement but haven't gotten to is the idea of a "public" or "guest" user that can have an even more restricted role than editor if desired. (maybe the guest can execute queries, but not modify them for example)

PR #646, opened 2020-05-31
Sessionless is admin or editor

eladeyal-intel:
Here's my attempt at solving #645.

So this changes legacy behavior in that that the default is for sessionless users is now "Editor" and not "Admin".

The "history" page is still available which is suboptimal.

### DIFF
--- a/config-example.ini
+++ b/config-example.ini
@@ -34,6 +34,10 @@ defaultConnectionId=''
 ; Set to TRUE to disable authentication altogether.
 disableAuth="false"
 
+; Specifies the role associated with users when disableAuth is set to true.
+; Acceptable values: admin, editor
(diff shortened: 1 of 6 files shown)


rickbergfalk:
What do you think about changing this configuration key to supply the role name to be used when auth is disabled? 

I'm not really sure what to call the setting, but it'd be nice if it corresponded with the setting that disables auth. 

Something like `disableAuthRole=editor`? Not sure it makes sense without having it explained though.

eladeyal-intel:
disableAuthDefaultRole ?

rickbergfalk:
> disableAuthDefaultRole ?

Sure!

eladeyal-intel:
ready for merge I think

PR #654, opened 2020-06-01
"trace" is not really a debug option

eladeyal-intel:
image

### DIFF
--- a/config-example.ini
+++ b/config-example.ini
@@ -115,10 +115,10 @@ timeoutSeconds=300
 ; Allows pre-approval of email domains. Delimit multiple domains by empty space.
 whitelistedDomains=""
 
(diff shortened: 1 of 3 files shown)


rickbergfalk:
According to pino docs it is a level, but the Logger wrapper isn't supporting it. 

https://github.com/pinojs/pino/blob/master/docs/api.md#loggertracemergingobject-message-interpolationvalues

Want to add trace option to https://github.com/rickbergfalk/sqlpad/blob/master/server/lib/logger.js#L4 and corresponding method similar to https://github.com/rickbergfalk/sqlpad/blob/master/server/lib/logger.js#L53

eladeyal-intel:
see #655

Issue #660, opened 2020-06-03
free anonymous users from connection selection


eladeyal-intel:
When logging anonymously (`disableAuth`), editor user is directed to `queries/new`. In my setup there is only one configured connection, but the is still presented with "..choose connection" where she must select the active connection in order to run.

There are two things which can be improved I think:

1. If there is only one connection available, do not show "...choose connection" (or even hide the connection dropbox at all) and auto select the connection.

2. allow the URL to connect the connection id - maybe `queries/new?connection=connectionname` and have this pre-configure the connection

rickbergfalk:
I'm open to both of these options. For specifying the connection in the url, I'd prefer either `connectionId` or `connectionName` fields be specified.

eladeyal-intel:
So the first option was merged :-) thank rick

I would love if someone helps implementing the second option, specifying connection name via URL

PR #662, opened 2020-06-05
Auto connection

eladeyal-intel:
Hide connection menu and auto-select if there's only one option - i.e. an editor with exactly 1 connection

### DIFF
--- a/client/src/queryEditor/ConnectionDropdown.js
+++ b/client/src/queryEditor/ConnectionDropdown.js
@@ -39,6 +39,11 @@ function ConnectionDropdown({
     connectConnectionClient();
   };
 
+  // Only show the connection menu if there's more than one option to select.
+  if (currentUser.role === 'editor' && connections.length === 1) {
+    return null;
+  }
+
   const style = !selectedConnectionId
     ? { color: '#777', width: 220 }
     : { width: 220 };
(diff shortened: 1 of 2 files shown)


eladeyal-intel:
Ready for review and merge I guess. Ref #660

rickbergfalk:
Can we put this in initApp instead? https://github.com/rickbergfalk/sqlpad/blob/master/client/src/stores/initApp.js

eladeyal-intel:
1. Even if we do move the selection to initApp, the decision to render or hide the ConnectionDropdown should stay in ConnectionDropdown.js right?

2. My initial proposal would have editors get the auto-selected connection; and admins would still get the "...select connection" since they can use the dropdown to select "New connection". If want to implement the same distinction while auto-selection in initApp.js, then information about the current user should be delivered into initApp.js. I don't know how to do it (and if it's possible).

3. We can auto-select automatically for admins as well, then it may be a bit confusing because the admin will see a selected connection but he can still use the dropdown for other activities.

eladeyal-intel:
I modified to auto select connection to all users

rickbergfalk:
Oh yeah. sorry my initial message wasn't clear. The auto-selecting can go into initApp and happen for all. For hiding the connection picker, I'm open to hiding it for everyone, or just non-admins. I thought it'd be nice to create connections from the drop down, but I realize too that it is maybe an anti-pattern and not a frequent action.

Issue #667, opened 2020-06-06
SequelizeTimeoutError on startup


eladeyal-intel:
I'm deploying with Docker, and finally trying to set a persistent volume under `/var/lib/docker` (so that container restart does not cause users data to be lost) .  The volume may be a bit slow - but the application certainly has read & write access.

sqlpad fails to start because of this, when trying to perform 

        CREATE TABLE IF NOT EXISTS `schema_version` (`name` VARCHAR(255) NOT NULL UNIQUE PRIMARY KEY);

I get
```
SequelizeTimeoutError: SQLITE_BUSY: database is locked
   at Query.formatError (/usr/app/node_modules/sequelize/lib/dialects/sqlite/query.js:419:16)
   at Query._handleQueryResponse (/usr/app/node_modules/sequelize/lib/dialects/sqlite/query.js:73:18)
   at Statement.afterExecute (/usr/app/node_modules/sequelize/lib/dialects/sqlite/query.js:250:31)
   at Statement.replacement (/usr/app/node_modules/sqlite3/lib/trace.js:19:31)
   at Statement.replacement (/usr/app/node_modules/sqlite3/lib/trace.js:19:31)","type":"Error","msg":"Error starting SQLPad"}
```

Any ideas? I thought there maybe there's a concurrency issue

The volume is CIFS / SMB.

eladeyal-intel:
I've read [here](https://www.sqlite.org/wal.html) that "WAL mode does not work over network. I thought that might be it but https://github.com/sequelize/sequelize/issues/5245 suggests the default is "DELETE" (and not WAL).

rickbergfalk:
Is Sequelize using WAL? I wasn't thinking it was. Is there anything else that might have a lock on the database? 

I'll have to take a look at the startup order of operations. Maybe there's something that is running asynchronously on startup that really should be sync.

rickbergfalk:
@eladeyal-intel weirdly enough I'm seeing this now in the migration test case. Investigating...

rickbergfalk:
Wait my mistake - I'm not seeing the BUSY error, but instead SQLITE_CANTOPEN: unable to open database file. Further investigation this CANTOPEN error seems to be a known issue https://github.com/sequelize/sequelize/issues/9691.

Going switch to using the old dialect/storage parameters for SQLite, otherwise use the URI if provided. 

Fixes my issue, not sure if it will help with this one.

eladeyal-intel:
[...]

PR #668, opened 2020-06-07
enable external DB for backend

eladeyal-intel:
I started this because of #667 and because I saw many docker apps doing the same and I think it would help scaling.

This PR adds configuration `SQLPAD_BACKEND_DB_URI`/`backendDatabaseUri` where one can specify an external database to be used for storing users, queries, etc, instead of sqlite.

This connecting string works for me

        SQLPAD_BACKEND_DB_URI=mariadb://user:password@hostname:port/dbname?debug=true&ssl=true

I added mariadb driver because it supports `ssl=true` on the connection string. Sequalize uses `mysql2` driver (if the connection string starts with `mysql://`) but I was not able to force ssl which is required in the environment I'm in.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -90,6 +90,13 @@ If enabled, runs embedded database `nedb` in memory. `dbPath` is still required 
 - Key: `dbInMemory`
 - Env: `SQLPAD_DB_IN_MEMORY`
 
+## backendDatabaseUri
+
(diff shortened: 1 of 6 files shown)


eladeyal-intel:
I think my options (in my environment) for persistent DB are only MariaDB and postgresql. I like MariaDB better for its verbose syntax.

I'll add an experimental note and try to add some testing.

eladeyal-intel:
@rickbergfalk I added "experimental" to documentation but the testing part seems too complex. 

I know you can change .travis.ci to run the test suite separately for each DB variant (see here https://github.com/frab/frab/blob/master/.travis.yml ) but it seems that the test suite will need to be changed to accommodate that; plus I'm not sure how to make linting run once only. (You can use "github actions" and then have different tasks for linting vs testing - but I don't know how to do it with travis-ci.org)

rickbergfalk:
Thanks @eladeyal-intel!

PR #670, opened 2020-06-07
lint in github actions

eladeyal-intel:
This is my attempt at moving the lint action from travis-ci.org to a dedicated github action. The benefit would be that the travis-ci will be able to run in parallel jobs (for tests only); and the linter rejects will be easy to detect on github GUI.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,22 @@
+
+name: lint
+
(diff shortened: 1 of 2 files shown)


rickbergfalk:
👍 I'm okay if Travis eventually gets replaced with GitHub actions too. Feel free to add more as you'd like 🙂

PR #676, opened 2020-06-09
hide old error message while testing

eladeyal-intel:
so now when user clicks "Test", the old error message is hidden while the test is being conducted.


rickbergfalk:
Thanks for being such a thorough follow up @eladeyal-intel! 😂 And I mean that in a serious way. Your attention to detail has been amazing and I appreciate all the misc fixes you've sent in :)

PR #679, opened 2020-06-12
Updates for SAML

eladeyal-intel:
1. The first commit in this PR allows renaming the "Sign in with ADFS" (used for SAML auth) with something else, i.e. "Sign in with NASA". The env variable `SAML_LINK_HTML` can include html so operator can add for example company (or a not-for-profit organization ) logo.

2. If I understand correctly, the SAML auth works only for users already added to the database. This PR adds env variable configuration `SQLPAD_SAML_AUTO_SIGN_UP` and `SQLPAD_SAML_DEFAULT_ROLE` which lets you admit all SAML users and assign roles.

### DIFF
--- a/client/src/SignIn.js
+++ b/client/src/SignIn.js
@@ -93,9 +93,15 @@ function SignIn({ config, refreshAppContext }) {
     </div>
   );
 
+  function createMarkupForSamlLink() {
+    return { __html: config.samlLinkHtml };
+  }
+
   const samlForm = (
     <div>
-      <a href={config.baseUrl + '/auth/saml'}>Sign in with ADFS</a>
+      <a href={config.baseUrl + '/auth/saml'}>
+        <span dangerouslySetInnerHTML={createMarkupForSamlLink()} />
+      </a>
     </div>
   );
 
--- a/server/routes/app.js
+++ b/server/routes/app.js
@@ -31,6 +31,7 @@ router.get(
         localAuthConfigured: !config.get('disableUserpassAuth'),
         publicUrl: config.get('publicUrl'),
         samlConfigured: Boolean(config.get('samlEntryPoint')),
+        samlLinkHtml: config.get('samlLinkHtml'),
         smtpConfigured: config.smtpConfigured(),
       },
       version: packageJson.version,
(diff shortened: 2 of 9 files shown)


server/routes/app.js
```diff
@@ -30,7 +30,8 @@ router.get(
         googleAuthConfigured: config.googleAuthConfigured(),
         localAuthConfigured: !config.get('disableUserpassAuth'),
         publicUrl: config.get('publicUrl'),
-        samlConfigured: Boolean(config.get('samlEntryPoint')),
+        samlConfig:
```
rickbergfalk, line 33: Could we leave in `samlConfigured` and send down `samlLinkHtml` as a separate, optional config field? I don't mind supporting custom link HTML, but I think it shouldn't be required. The default button today could be changed to something like "Sign in with SSO" or something generic to fix the ADFS  naming

eladeyal-intel:
image


```
samlLinkHtml="<IMG style='height: 1.2rem;vertical-align:text-bottom;' src='http://seekvectorlogo.com/wp-content/uploads/2018/02/nasa-vector-logo-small.png'/> Sign in with NASA"
```

client/src/SignIn.js
```diff
@@ -93,9 +93,15 @@ function SignIn({ config, refreshAppContext }) {
...
 
+  function createMarkupForSamlLink() {
+    return { __html: config.samlLinkHtml };
+  }
+
   const samlForm = (
     <div>
-      <a href={config.baseUrl + '/auth/saml'}>Sign in with ADFS</a>
+      <a href={config.baseUrl + '/auth/saml'}>
+        <span dangerouslySetInnerHTML={createMarkupForSamlLink()} />
```
rickbergfalk, line 103: Can we either provide a default value for samlLinkHtml in config or a default button here if HTML is not provided?

eladeyal-intel: I have a default value in config-items.js. It's not enough?

rickbergfalk: Ah sorry I had missed the text. I was skimming for something that looked like HTML - this works

PR #681, opened 2020-06-13
allow connectionName, connectionId on URL

### LINKED ISSUE #660: free anonymous users from connection selection

eladeyal-intel:
closes #660

I'm not sure where it makes sense to document this URL

example usage: 
- `http://localhost:3000/sqlpad/queries/new?connectionName=xxx`
- `http://localhost:3000/sqlpad/queries/new?connectionId=xxx-xx-xxx-xx-xxxxx`

### DIFF
--- a/client/src/stores/initApp.js
+++ b/client/src/stores/initApp.js
@@ -3,6 +3,7 @@ import message from '../common/message';
 import { refreshAppContext } from './config';
 import fetchJson from '../utilities/fetch-json';
 import sortBy from 'lodash/sortBy';
+const queryString = require('query-string');
 
 window.localforage = localforage;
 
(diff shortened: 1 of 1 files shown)


client/src/stores/initApp.js
```diff
@@ -56,6 +57,17 @@ const initApp = async (state) => {
           update.selectedConnectionId = selectedConnectionId;
         }
       }
+
+      const qsConnectionName = queryString.parse(window.location.search)[
```
rickbergfalk, line 61: Could we use the dot notation here for access? `queryString.parse(window.location.search).connectionName` should work.

client/src/stores/initApp.js
```diff
@@ -56,6 +57,17 @@ const initApp = async (state) => {
...
+        'connectionName'
+      ];
+      if (qsConnectionName) {
```
rickbergfalk, line 64: Could this also be updated to support `connectionId` as a parameter?

eladeyal-intel:
@rickbergfalk - I'm getting increasing worried that the term `_id` in the file AppInit should actually be `id`.

client/src/stores/initApp.js
eladeyal-intel: done

rickbergfalk:
> @rickbergfalk - I'm getting increasing worried that the term `_id` in the file AppInit should actually be `id`.

Ah good catch - `_id` was renamed to `id` in the v5 work to move to SQLite.

rickbergfalk:
Thanks for this @eladeyal-intel!

rickbergfalk:
I forgot to add - I don't know where this makes sense to document either. Maybe we can add a page to the API documentation for query string parameters or something?

rickbergfalk:
@eladeyal-intel just heads up I'll look into and fix `_id` issue quick. I'm working on trying to get the state store stuff in better shape and I'm in the middle of refactoring initApp

Issue #647, opened 2020-05-31
external tables


acivitillo:
It seems that external tables are not supported. We are using Redshift Spectrum and its external tables can be queried in SqlPad but they don't show up in the tree left sidebar.

rickbergfalk:
Does Redshift use the postgres driver to connect? 

Its likely that the schema query would need to be altered to get at external tables. I'm curious if external tables are also missing from the postgres connections as well.

If the query is not compatible, we can create a redshift specific driver option in SQLPad (that still uses the postgres driver behind the scenes) and runs a different query designed for redshift schema.

acivitillo:
Correct, same driver, but this might be a Redshift specific query: https://docs.aws.amazon.com/redshift/latest/dg/r_SVV_EXTERNAL_TABLES.html

Your approach should be correct because the query will pass through the postgres driver but it is a different query, this is what we need to query to get the list of tables:

```sql
SELECT *
FROM SVV_EXTERNAL_TABLES
WHERE schemaname='spectrum'
```

acivitillo:
I see, so we'd have to change this: https://github.com/rickbergfalk/sqlpad/blob/master/server/drivers/postgres/index.js

And the below query.

I can give you the query if it helps? I am not that good with javascript and wouldn't know how to replicate your coding environment on my local PC to add the changes and the tests. I can do it but it will take few weeks, let me know.

Meanwhile as for the query, I guess we'd need to do a UNION if we want to stick to 1 SCHEMA_SQL. We'd union external and normal tables.

Of course the best solution would be for the sidebar to become a tree view and show tables, external and views nodes. But for now just throwing all tables and external tables under a schema would be great. 

```
const SCHEMA_SQL = `
  select 
    ns.nspname as table_schema, 
    cls.relname as table_name, 
    attr.attname as column_name,
    trim(leading '_' from tp.typname) as data_type,
    pg_catalog.col_description(attr.attrelid, attr.attnum) as column_description
  from 
    pg_catalog.pg_attribute as attr
    join pg_catalog.pg_class as cls on cls.oid = attr.attrelid
    join pg_catalog.pg_namespace as ns on ns.oid = cls.relnamespace
    join pg_catalog.pg_type as tp on tp.typelem = attr.atttypid
  where 
    cls.relkind in ('r', 'v', 'm')
    and ns.nspname not in ('pg_catalog', 'pg_toast', 'information_schema')
    and not attr.attisdropped 
    and attr.attnum > 0
[...]

PR #698, opened 2020-06-20
Rename Redshift Spectrum to Redshift

rickbergfalk:
Renames Redshift Spectrum to Redshift. 

If a previously existing redshift connection is set via config or seed data, its `driver` field must be changed to `redshift` instead of `redshiftSpectrum`


rickbergfalk:
This is also removes the "experimental" label as well. 

Fixes #647

PR #708, opened 2020-06-25
don't prompt for LDAP if not configured

eladeyal-intel:
Users should not see the word LDAP if LDAP authentications is not configured in.

Well actually - users should not see the word LDAP ever - so I just used "username"


rickbergfalk:
@eladeyal-intel I'll merge this one in for now since the tests aren't running. Hopefully they will going forward

PR #710, opened 2020-06-25
Using STRING instead of JSON to support MSSQL dialect in SQLPAD_BACKEND_DB_URI

yorek:
Currently Sequelize do not support JSON data type for MSSQL dialect. In SQL Server / Azure SQL, JSON is just a string. This change will allow SQLPAD to store metadata into an Azure SQL database. This is a temporary fix until Sequelize will properly support JSON for MSSQL (I'm actively working on it, but it will take some weeks). 

As JSON is used just to store data a no query are done on it, this should make no harm to anyone.


yorek:
Forgot to sync with upstream before submitting a PR. Fixed, by syncing later, so only the three changed files are now correctly mentioned in the "Files Changed" panel.

eladeyal-intel:
I'm not sure I understand what's going on there. But if you are to change a persistant database,  Ithink you'd want to add this as a new migration and not modify historic migrations, which (I think) would break existing deployments. See https://github.com/sequelize/umzug .

yorek:
The problem is that the moment Sequelize will try to run something like 
```
CREATE TABLE table (column JSON)
```
it will fail on SQL Server / Azure SQL, and I don't think that can be fixed just using a migration script....as the issue comes from the migration script itself.

eladeyal-intel:
I am familiar with other migration systems, where this could be solved because building a fresh DB is not necessarily going though each of the migrations. 

In my deployment (mariadb) the JSON appears as "longtext" - so this change may work or not work. 

I can imagine that sqlite is not affected by this change. Needs to be tested.

rickbergfalk:
I can help in writing some tests, but I think changing this from JSON to string has other impacts too. I thought the JSON sequelize type would stringify/parse the column value automatically when set to JSON too...

As for how to implement this we might need a combination of both adding a migration to modify existing databases in combination of changing the existing migrations. 

At some point I plan grouping all the migrations up as a base. That won't be for a while yet though

yorek:
Thank for the feedback. I have found a workaround, that would solve the JSON data type issue. There is still a small fix to do as there is an issue with a view. But I'll open a dedicated PR for that.

Issue #711, opened 2020-06-25
Cannot build docker image


yorek:
Docker image cannot be built as the following error pops up:

```
api/batches
    ✓ Creates batch
    1) GETs finished result


  3 passing (11s)
  1 failing

  1) api/batches
       GETs finished result:
     Error: Timeout of 10000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/sqlpad/server/test/api/batches.js)
      at listOnTimeout (internal/timers.js:549:17)
      at processTimers (internal/timers.js:492:7)
```

and after that, the build process stops.

I'm using Windows 10 WSL 2 (32GB, 8vCores)

rickbergfalk:
@yorek how are you building the image?

I'm on Windows 10 with WSL 2 as well, but things work for me. I run docker commands using cmder, from Windows as opposed to from bash inside WSL 2 if that makes sense? I switched back to Windows a few months ago and I'm still sometimes confused about what is running where.

rickbergfalk:
@yorek I wonder if something is running into an error and it isn't obvious because the logs are set to silent? 

I've pushed changes to `server/test/utils.js` to log error level messages. Hopefully next test run we'll find something pointing to the issue.

yorek:
Will do. I tried both running the test using WSL or Powershell....and it doesn't make any difference. This is expected as when Docker starts everything happens inside the Docker container so no matter how you start it, the result should be the same

yorek:
Btw I'm building the image using a very simple
```
docker build . -t sqlpad
```

yorek:
Ok, after synching by fork with the upstream repo I'm now able to build the image. I also fixed an issue with the view I changed.
Unfortunately now I cannot push my changes due to #719

yorek:
Fixed (see #719)

PR #715, opened 2020-06-25
make view compatible with MSSQL

yorek:
The original 
```
MAX(incomplete) AS incomplete
```
doesn't work on Azure SQL / SQL Server. It generates an exception and the migrations will stop. Therefore this change cannot be done as an additional migration step. Luckily CONVERT is ANSI-SQL it should not break anything on MYSQL or PostgreSQL


yorek:
Would love to test this via the provided tests, but I'm having this issue #711 that at the moment prevents me to run them.

rickbergfalk:
@yorek I think if you pull latest master into your branch, and then push again the tests will run via GitHub actions. Would still like to get your setup up and running but at least you can look for tests run here on GitHub.

server/migrations/04-00400-query-history-to-batch.js
```diff
@@ -86,7 +86,7 @@ async function up(queryInterface, config, appLog, nedb, sequelizeDb) {
           SELECT 
             batch_id, 
             SUM(row_count) AS row_count, 
-            MAX(incomplete) AS incomplete
+            MAX(CONVERT(INT, incomplete)) AS incomplete
```
rickbergfalk, line 89: We'll also need to add a new migration that does this same thing but updates the view.

Changing this migration will work for new instances, but existing installations won't have this change since the migration will have already been recorded as "run" and ignored.

Unsure if there's an ANSI SQL create or replace view... we may need to provide different implementations based on backend db uri.

I can handle the new migration once I get my mssql test setup working in https://github.com/rickbergfalk/sqlpad/pull/717

rickbergfalk:
I forgot about `server/routes/queries.js` which builds up a SQL string... and uses the LIMIT keyword :(

There are probably a few places where this is incompatible with various backend dbs. Not a lot... but it'll require working through.

I'm open to pushing through and figuring this out though. Despite adopting Sequelize, I'm not really a big fan of it. For anything beyond the basic CRUD operations on a table we should maybe look at adopting knex to build SQL strings.

yorek:
It seems that Sequelize would take care of executing the corrent OFFSET/LIMIT or OFFSET/FETCH command, depending on the dialect: https://sequelize.org/v5/manual/querying.html

server/migrations/04-00400-query-history-to-batch.js
yorek: Correct, but existing installations won't have this problem at all, as they were not done on MSSQL :) I can add a new migration that will drop and recreate the view with the new code, so that everyone will be on the same page. Still the fix to the existing migration needs to be there otherwise for MSSQL SqlPad won't run if using MSSQL for storing metadata

PR #717, opened 2020-06-26
Potential backendDatabaseUri testing strategy

rickbergfalk:
@yorek @eladeyal-intel Out of curiosity I wanted to see what it'd take to get a testing set up in place where we could run the suite of tests against a targeted backend. 

This is quickly put together, but what I've come up with is to set an env var to key off of, and then in test utils dynamically create and use databases on the targeted backend. All the tests today assume they can start fresh with a database, which is easy to do with SQLite.

Here's what I've got so far with this:

* In one terminal window, cd to `/server` and do `docker-compose up` to get sql server running
* In another terminal (or in same if you run previous command in background) do `npm run test-mssql`, and a test databases should be created in test utils.

Doing this I immediately hit a failing test 

```
 1) api/app
       "before all" hook for "returns expected values":
     SequelizeDatabaseError: Column, parameter, or variable #5: Cannot find data type JSON.
```

Is that what you get @yorek?

Feel free to use and commit to this branch if its helpful.

### DIFF
--- a/server/test/utils.js
+++ b/server/test/utils.js
@@ -15,6 +15,7 @@ const ensureConnectionAccess = require('../lib/ensure-connection-access');
 
 const USE_MSSQL = process.env.SQLPAD_TEST_DB === 'mssql';
 const USE_PG = process.env.SQLPAD_TEST_DB === 'pg';
+const USE_MYSQL = process.env.SQLPAD_TEST_DB === 'mysql';
 
 // At the start of any test run, clean out the root artifacts directory
 before(function (done) {
@@ -33,6 +34,8 @@ class TestUtils {
       backendDatabaseUri = `mssql://sa:SuperP4ssw0rd!@localhost:1433/${this.dbname}`;
     } else if (USE_PG) {
       backendDatabaseUri = `postgres://sqlpad:sqlpad@localhost:5432/${this.dbname}`;
+    } else if (USE_MSSQL) {
+      backendDatabaseUri = `mysql://sqlpad:sqlpad@localhost:3306/${this.dbname}`;
     }
 
     const config = new Config(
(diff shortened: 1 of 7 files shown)


rickbergfalk:
Alright so I had to try postgres out to see if this approach works, and after fixing a few tests it does! 

@eladeyal-intel look what you started 😂

eladeyal-intel:
I have it deployed here with mariadb and it seems to work fine. Maybe it will be fast to enable test for

rickbergfalk:
@eladeyal-intel going to do that now 👍

rickbergfalk:
Something I'm finding interesting... and I kind of already knew this but ignored it... MySQL, SQL Server, and Postgres are a lot more strict than SQLite. Pointing tests at other databases are finding issues that should have come up with SQLite, but it is so forgiving that things kinda worked anyways.

Not sure if I like that or not.

rickbergfalk:
@eladeyal-intel mysql required no changes 👍

yorek:
> Doing this I immediately hit a failing test
> 
> ```
>  1) api/app
>        "before all" hook for "returns expected values":
>      SequelizeDatabaseError: Column, parameter, or variable #5: Cannot find data type JSON.
> ```
> 
> Is that what you get @yorek?

Nope, with MSSQL this is expected. A way to solve this is to create a JSON data type: 

```
create type [dbo].[JSON] from [nvarchar](max) null
```

that would solve the JSON data type issue. Problem with the view I mentioned in  #715 will still be there.

yorek:
#715 fixed now

server/test/utils.js
```diff
@@ -80,6 +101,28 @@ class TestUtils {
...
 
   async initDbs() {
+    // If mssql is given we need to create a db first
+    const backendDatabaseUri = this.config.get('backendDatabaseUri') || '';
+    if (backendDatabaseUri.startsWith('mssql')) {
+      const masterUri = backendDatabaseUri.replace(this.dbname, 'master');
+      const sequelize = new Sequelize(masterUri, {
+        logging: (message) => appLog.debug(message),
+      });
+      await sequelize.query(`CREATE DATABASE ${this.dbname};`);
```
yorek, line 111: after this just add
```
const sequelize2 = new Sequelize(backendDatabaseUri, {
        logging: (message) => appLog.debug(message),
      });
await sequelize2.query(`CREATE TYE [dbo].[JSON] FROM [NVARCHAR](MAX) NULL;`);
```
and that should do the trick

rickbergfalk:
Realized last night that my mysql test wasn't actually testing mysql. 

Once that was fixed, learned that  the `mariadb` URI works, but `mysql` does not due to the way JSON is handled. JSON fields using `mariadb` are returned parsed as an object, where `mysql` returns JSON fields as string.

rickbergfalk:
Alright actually could be my fault with trying to use `mysql` URI against a mariadb database. https://github.com/sequelize/sequelize/issues/10946

Will need to separate mysql and mariadb tests

PR #720, opened 2020-06-26
tests for multiple SQLPAD_BACKEND_DB_URI (partial)

eladeyal-intel:
@rickbergfalk , I've done this based on your work in #717 - you can see that it invokes several jobs for the various backeds databases. I was not have them pass. A couple of tests I manipulated just to get things going - not sure what they were supposed to be testing.


rickbergfalk:
Thanks @eladeyal-intel! Looks like there will be subtle changes needed for tests around how dates are handled. Will merge and will gradually work at getting things compatible.

PR #723, opened 2020-06-27
multiple backends testing improved

eladeyal-intel:
1. Now things are built once, and the built artifact is used for multiple tests (which run in parallel)

2. Added tests, we have now mysql, mariadb, mssql, psql and sqlite. All but the last fail (but for interesting reasons)

3. Servers are no longer launched unnecessarily - so the mssql test will launch the mssql docker only.

3. I moved all the images into the server/docker-compose.yaml and they're not in the github action. This way we can reuse the same for dev. For script convenience , I changed yours "test-mysql" to "testmysql"

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
-      - name: Build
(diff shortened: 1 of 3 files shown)


rickbergfalk:
Will see if I can get some of these compatible today

PR #729, opened 2020-06-27
fix mssql json type for test

eladeyal-intel:
So this implements @yorek comment on  https://github.com/rickbergfalk/sqlpad/pull/717 (earlier I've tried something different). It does not seem to "do the trick". I get a `SequelizeUniqueConstraintError: Validation error`. Maybe @yorek wants to take it from here

### DIFF
--- a/server/test/utils.js
+++ b/server/test/utils.js
@@ -105,8 +105,11 @@ class TestUtils {
       }
 
       if (backendDatabaseUri.startsWith('mssql:')) {
-        await sequelize.query(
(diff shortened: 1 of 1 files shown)


rickbergfalk:
same in #730 😂

rickbergfalk:
Figured out the constraint error -- It is the nullable unique constraint on password reset id. For SQL Server this needs to be enforced on non-null values. https://github.com/rickbergfalk/sqlpad/blob/master/server/migrations/04-00200-nedb-sqlite-tables.js#L331

PR #733, opened 2020-06-27
faster tests

eladeyal-intel:
This is supposed to make CI tests a bit faster, since some of the service bring-up is done in parallel to other tasks.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -45,6 +45,10 @@ jobs:
         with:
           node-version: ${{ matrix.node-version }}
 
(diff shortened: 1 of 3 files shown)


rickbergfalk:
@eladeyal-intel been noticing mysql has been losing connection at times... not sure what is up with that.

```
1) api/app
       "before all" hook for "returns expected values":
     SequelizeConnectionError: Connection lost: The server closed the connection.
      at ConnectionManager.connect (node_modules/sequelize/lib/dialects/mysql/connection-manager.js:126:17)
```

PR #734, opened 2020-06-27
Fix MSSQL JSON and queries list API

rickbergfalk:
Adds work around for lack of proper JSON support in SQL Server. 

While the JSON datatype can be added to SQL Server as a string, Sequelize does not parse that string to an object. As a solution, the field checked and parsed in application code after the fact. 

This isn't great, but it works for now.

This also fixes a SQL incompatibility in the queries list API route to address the fact that SQL Server does not support LIMIT/OFFSET.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,6 @@
 name: test
 
-on: [pull_request]
+on: [push, pull_request]
 
 jobs:
   build:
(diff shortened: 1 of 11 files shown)


rickbergfalk:
@eladeyal-intel I'm seeing duplicate actions being run on my end for both push and pull_request triggers. 

Looks like others hit this https://github.community/t/duplicate-checks-on-push-and-pull-request-simultaneous-event/18012/2

Do you see any issue changing this to pull_request only or potentially adding the if-check on repo name?

eladeyal-intel:
I'm not very exipirnces with github actions. In #736 I tried to do what you suggested - let's see if it helps.

rickbergfalk:
@eladeyal-intel yeah I was thinking about that and the switch to PR only.

rickbergfalk:
I think mssql might be green now 🤞

eladeyal-intel:
👏

PR #743, opened 2020-06-28
Create JSON datatype for MSSQL

yorek:
Moved the logic of creating the JSON data type in the first migration script, only in case MSSQL is used


rickbergfalk:
@yorek will let you do the honors of merging :)

PR #756, opened 2020-07-02
Retry on bigquery network timeouts

bruth:
The nodejs-bigquery SDK has a hard-coded 60s fetch timeout (see https://github.com/googleapis/nodejs-bigquery/blob/13b7e391cb3cfd87caec094f058143842cb39306/src/table.ts#L1909). This results in the query timing out at the network level even if the SQLPad and BigQuery API timeouts are set higher.

The solution is to simply retry the getQueryResults ignoring network timeout errors.

Signed-off-by: Byron Ruth <b@devel.io>

### DIFF
--- a/server/drivers/bigquery/index.js
+++ b/server/drivers/bigquery/index.js
@@ -6,6 +6,36 @@ const id = 'bigquery';
 const name = 'BigQuery';
 
 /**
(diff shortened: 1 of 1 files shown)


rickbergfalk:
Thanks @bruth!

PR #769, opened 2020-07-06
env var consistency GOOGLE

eladeyal-intel:


eladeyal-intel:
Note: I did not test this..

PR #773, opened 2020-07-07
Use ENV for docker image defaults

### LINKED ISSUE #771: Overwrite Docker port?
https://github.com/rickbergfalk/sqlpad/blob/master/docker-entrypoint#L2
Looking at this entry point, it is using the CLI variable to set the port to 3000

Then diving into the config code https://github.com/rickbergfalk/sqlpad/blob/master/server/lib/config/index.js#L24 it looks like the CLI variable wins if there are conflicts between ENV, File, and CLI.  

Am I reading this correctly?  If so, then without changing the startup script there is no way to change the startup port.  

In normal Docker, I could map 3000 within the Docker network.  However, I'm looking to run this in a public cloud provider and they don't support the port mapping.  Thoughts?

rickbergfalk:
Use ENV for docker image defaults. This allows docker port to be overridden 

fixes #771

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -54,6 +54,8 @@ COPY --from=build /sqlpad/docker-entrypoint /
 COPY --from=build /sqlpad/server .
 
 ENV NODE_ENV production
+ENV SQLPAD_DB_PATH /var/lib/sqlpad
+ENV SQLPAD_PORT 3000
 EXPOSE 3000
 ENTRYPOINT ["/docker-entrypoint"]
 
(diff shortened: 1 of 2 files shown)


eladeyal-intel:
Should the EXPOSE command in docker file be edited to use the SQLPAD_PORT then?

rickbergfalk:
I'm not sure it matters. From documentation EXPOSE serves as documentation of sorts and for automation, but the port can be overridden. Even with EXPOSE 3000, I was able to change SQLPAD_PORT to 3001 and map it to my host.

PR #779, opened 2020-07-10
Rename and remove deprecation for HTTPS environment variables

rickbergfalk:
* Renames HTTPS related environment variables to be prefixed with `SQLPAD_HTTPS_*`. 
* Removes deprecation notice, instead providing a note about reverse proxy in config documentation


rickbergfalk:
cc @eladeyal-intel

PR #787, opened 2020-07-11
show service token UI only if enabled in config

eladeyal-intel:


server/routes/app.js
```diff
@@ -41,6 +41,8 @@ async function getApp(req, res) {
         config.get('ldapAuthEnabled') || config.get('enableLdapAuth'),
       oidcConfigured: config.oidcConfigured(),
       oidcLinkHtml: config.get('oidcLinkHtml'),
+      showServiceTokensUI:
+        config.get('serviceTokenSecret') || config.get('serviceTokenSecret_d'),
```
rickbergfalk, line 45: Can you wrap this with `Boolean(config.get('serviceTokenSecret') || config.get('serviceTokenSecret_d'))`. Secret would leak to UI otherwise

eladeyal-intel: @rickbergfalk done! nice catch

PR #793, opened 2020-07-15
CrateDB - UI changes - added user,pwd,ssl field

### LINKED ISSUE #791: CrateDB Connection - Add User/Password Authentication + SSL
Right now it is possible to connect to a CrateDB Cluster with SSL and username/password, when adding the whole connection string to the "hostname" field: ```https://USER:PWD@host.cratedb.net:4200```

To hide the password and for convenience, this should be separated into 4 fields and a checkbox:
- hostname
- port
- username
- password
- ssl checkbox

proddata:
Closes #791 

- Added fields for
  - username
  - password
  - ssl toggle
- Added connection string builder
- Added CrateDB default port (4200)

- [x] ran npm lint
- [x]  ran tests

### DIFF
--- a/server/drivers/crate/index.js
+++ b/server/drivers/crate/index.js
@@ -3,7 +3,6 @@ const { formatSchemaQueryResults } = require('../utils');
 
 const id = 'crate';
 const name = 'Crate';
-const CRATE_DEFAULTPORT = '4200';
 
 // NOTE per crate docs: If a client using the HTTP or Transport protocol is used a default limit of 10000 is implicitly added.
 // node-crate uses the REST API, so it is assumed this is a limit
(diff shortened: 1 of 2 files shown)


server/drivers/crate/index.js
```diff
@@ -69,6 +67,26 @@ async function runQuery(query, connection) {
   }
 }
 
+/**
+ * Get connection string with options
+ * @param {*} connection
+ */
+function getConnectionString(connection) {
+  let connectionString = connection.ssl ? 'https://' : 'http://';
+  if (connection.username !== undefined && connection.username.length !== 0) {
```
rickbergfalk, line 76: I'm not sure if it is possible, but `null` could potentially be passed in and would break without checking for it. 

Can this check be changed to either `if (connection.username)`, or `if (typeof connection.username === 'string' && connection.username.length !== 0)`?

`null`, `undefined` and empty string`''` all evaluate as falsy

server/drivers/crate/index.js
```diff
@@ -69,6 +67,26 @@ async function runQuery(query, connection) {
...
+    connectionString += connection.username;
+    if (connection.password !== undefined && connection.password.length !== 0)
+      connectionString += ':' + connection.password;
+    connectionString += '@';
+  }
+  connectionString += connection.host + ':';
+  connectionString +=
+    connection.port === undefined || connection.port.length === 0
```
rickbergfalk, line 84: similar challenge here as well, but we should handle string or number for port. For this I'd go with 

```js
connectionString += Boolean(connection.port) ? CRATE_DEFAULTPORT : connection.port;
```

server/drivers/crate/index.js
```diff
@@ -69,6 +67,26 @@ async function runQuery(query, connection) {
   }
 }
 
+/**
+ * Get connection string with options
+ * @param {*} connection
+ */
+function getConnectionString(connection) {
+  let connectionString = connection.ssl ? 'https://' : 'http://';
```
rickbergfalk, line 75: Should this check if protocol is supplied in current host field?

I'm wondering if anyone may have included protocol previously, and this would double up on it or not.

If protocol is supported in previous implementation, perhaps host is parsed with URL first? https://nodejs.org/docs/latest-v12.x/api/url.html

server/drivers/crate/index.js
proddata: I changed it to `if (connection.username)`

server/drivers/crate/index.js
proddata: the key `port` is defined as text, shouldn't I expect to only handle text?
changed it to `if (connection.port)`
Also I removed the default port, after checking `node-crate` again

server/drivers/crate/index.js
proddata: Thanks for the input!
I reworked it to use URL in general. This should take care of case where a combination of `host`, `port`, `auth` or `protocol` have previously been supplied in the `host` field.

rickbergfalk: Excellent! I wasn't aware of URL myself until not long ago.

rickbergfalk:
Thanks @proddata!

Issue #810, opened 2020-07-28
mysql connections lost; treated as fatal


eladeyal-intel:
reproducable in v5.2.1

I start an MySQL connection with transaction support. The server is MariaDB and the sqlpad driver in mysql
run a query.
It seems that after a predefined time the session/ connection is closed automatically. (In my case it seems to be after 2 minutes)

```json
   2020-07-28T11:33:42.03+0300 [APP/PROC/WEB/0] OUT {"level":50,"time":"2020-07-28T08:33:42.035Z","pid":7,"hostname":"ed954926-3567-4772-455c-19fc","name":"sqlpad-app","fatal":true,"code":
"PROTOCOL_CONNECTION_LOST","stack":"Error: Connection lost: The server closed the connection.\n    at Protocol.end (/usr/app/node_modules/mysql/lib/protocol/Protocol.js:112:13)\n    at TLSSocket.onend
(_stream_readable.js:679:10)\n    at Object.onceWrapper (events.js:417:28)\n    at TLSSocket.emit (events.js:323:22)\n    at endReadableNT (_stream_readable.js:1204:12)\n    at processTicksAndRejections
(internal/process/task_queues.js:84:21)","type":"Error","msg":"Connection lost: The server closed the connection."}
```

moreover, trying another query, I get (on the GUI) `Cannot enqueue Query after fatal error.` The only way to get rid of this is to reselect the conenction from the connection list or refresh the screen.

I would expect the connection to be closed so soon. 

I would definitely expect  a new connection to be created in this case.

I think I would like some visual feedback for "new connection created". Maybe related to #806 - the client can remember which connection was used. If it was dismissed by user action - drop response silently. If the connection was replaced because of server side decision, display a notification somewhere.

eladeyal-intel:
- mysql2 seems to do the same, but a different error message is displayed

- another workaround is to disconnect + re-connect using this icon:
![image](https://user-images.githubusercontent.com/20379005/88660552-9be54600-d0df-11ea-8844-44e1ca4979e2.png)

rickbergfalk:
I think what needs to happen here is the mysql databse connection needs to be used continuously to prevent it from closing? 

For example maybe a `SELECT 1` is sent every min or so to tell MySQL not to close the connection.

eladeyal-intel:
I've seen somewhere that js sql driver supports ping() for keep alive.

I wonder if what the mysql cli do to avoid being disconnected.

[...]

PR #813, opened 2020-08-01
mysql connections are interactive (#810)

eladeyal-intel:
INTERACTIVE - Indicates to the MySQL server this is an "interactive" client. This will use the interactive timeouts on the MySQL server and report as interactive in the process list. (Default off)

https://www.npmjs.com/package/mysql/v/2.18.1


eladeyal-intel:
I didn't test that if it makes any difference

PR #837, opened 2020-09-13
LDAP Auto Signup and RBAC

skaravad:
Based on my limited nodejs knowledge, I think I got it working, please review. I've tested it locally and it works.

### DIFF
--- a/server/auth-strategies/ldap.js
+++ b/server/auth-strategies/ldap.js
@@ -28,14 +28,23 @@ function enableLdap(config) {
           bindCredentials:
             config.get('ldapPassword') || config.get('ldapPassword_d'),
           searchFilter: config.get('ldapSearchFilter'),
+         // Enable Group Search, very  limited RBAC
+         // searchfilter should include the group
+         // SQLPAD_LDAP_SEARCH_FILTER='(&(uid={{username}})(memberOf=<GROUP_DN>))'
+          groupSearchBase: config.get('ldapBaseDN'),
+          groupSearchFilter: '(cn={{dn}})',
         },
       },
       async function passportLdapStrategyHandler(req, profile, done) {
         try {
           const { models } = req;
-          const mail = profile.mail.toLowerCase();
+          const email = profile.mail.toLowerCase();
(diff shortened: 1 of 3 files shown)


server/auth-strategies/ldap.js
```diff
@@ -28,26 +28,59 @@ function enableLdap(config) {
...
+          // get all groups that user belongs to, not sure how to substitue memberOf from env var
+          const groups = profile.memberOf;
+          // match the groups which are predefined, refer to configuration
+          if (groups.includes(admin_group)) {
+            appLog.debug(`${uid} successfully logged in with role admin`);
+            global.role = 'admin';
+          } else if (groups.includes(editor_group)) {
+            appLog.debug(`${uid} successfully logged in with role editor`);
+            global.role = 'editor';
+          }
```
rickbergfalk, line 52: Instead of using `global.role`, could we use a local `let role;` instead? 

`let` allows changing the value assignment in JS. For example you can do

```js
let role;
if (groups.includes(admin_group)) {
  role = 'admin';
} else if (groups.includes(editor_group)) {
  role = 'editor';
}
```

You may get a lint warning later, since the variable name is the same name as the key. This is fine but there's a shorthand you can use. 

This:
```js
const newUser = await models.users.update(user.id, {
  role: role,
});
```

can become this (as long as the local variable `role` is in scope)

```js
const newUser = await models.users.update(user.id, {
  role,
});
```

rickbergfalk (changes requested):
Sorry for the delay in reviewing this. This looks great @skaravad! 

A minor change around using `global`. For not being familiar with JS you did great 👍

server/auth-strategies/ldap.js
skaravad: Thank you for the feedback @rickbergfalk , I'll make the changes , will need to test it though :) will try to find some time this week.

rickbergfalk:
@skaravad Hope you don't mind but I took this in and worked it a bit. New PR is #857 based on your work in this PR :)

PR #857, opened 2020-10-21
LDAP Auto Signup and RBAC - Take 2

rickbergfalk:
I took at pass at updating and testing out #837, which led me down a path of running a local LDAP server via this docker image https://github.com/rroemhild/docker-test-openldap, and updating the implementation to be able to handle that. 

I'm not too familiar with LDAP, but from what I understand `memberOf` isn't standard, and might not always be available? 

With that in mind, I changed the config variables a bit to be more generic from the original implementation. To get dynamic role assignment via LDAP, you must set an attribute you want to look to, and values for each role. 

For example, assume the LDAP profile had a field called `employeeType`, with values `Manager` and `Individual contributor`, you could do something like the following:

```sh
SQLPAD_LDAP_ROLE_ATTRIBUTE="employeeType"
SQLPAD_LDAP_ROLE_ADMIN_VALUE="Manager"
SQLPAD_LDAP_ROLE_EDITOR_VALUE="Individual contributor"
```

Other config introduced here are auto sign up and default role. Default role is used if the role attribute/values missed

```sh
SQLPAD_LDAP_AUTO_SIGN_UP=true
SQLPAD_LDAP_DEFAULT_ROLE = "editor"
```

Also fixed in this PR - LDAP can be enabled and local auth disabled! I can't remember where this was at or what the plan was, but I do recall it being confusing.

@dengc367 @eladeyal-intel @skaravad any thoughts since you all are more familiar with LDAP than myself?

### DIFF
--- a/server/auth-strategies/ldap.js
+++ b/server/auth-strategies/ldap.js
@@ -40,7 +40,6 @@ function queryLdap(client, searchBase, scope, filter) {
         return reject(err);
       }
 
-      // eslint-disable-next-line no-unused-vars
       res.on('searchEntry', function (entry) {
         results.push(entry.object);
       });
--- a/server/auth-strategies/ldap.js
+++ b/server/auth-strategies/ldap.js
@@ -93,6 +93,8 @@ function enableLdap(config) {
           const adminRoleFilter = config.get('ldapRoleAdminFilter');
           const editorRoleFilter = config.get('ldapRoleEditorFilter');
 
+          // At least with test setup, jpegPhoto is gnarly output
+          // Remove prior to logging
           delete profile.jpegPhoto;
           appLog.debug(profile, 'Found user');
 
@@ -104,6 +106,7 @@ function enableLdap(config) {
           }
 
           // Derive a userId fiter based on profile that is found
+          // ActiveDirectory will have sAMAccountName, while OpenLDAP will have uid
           let userIdFilter = '';
           if (profile.sAMAccountName) {
             userIdFilter = `(sAMAccountName=${profile.sAMAccountName})`;
@@ -215,6 +218,10 @@ function enableLdap(config) {
             });
             return done(null, newUser);
           }
+
+          // If no user was found, and config does not allow initial log in or auto-creating users,
+          // Return not authorized
+          return done(null, false);
         } catch (error) {
           done(error);
         }
--- a/server/auth-strategies/ldap.js
+++ b/server/auth-strategies/ldap.js
@@ -72,6 +72,9 @@ function enableLdap(config) {
     config.get('ldapBaseDN') ||
     config.get('ldapBaseDN_d');
 
+  const adminRoleFilter = config.get('ldapRoleAdminFilter');
+  const editorRoleFilter = config.get('ldapRoleEditorFilter');
+
   appLog.info('Enabling ldap authentication strategy.');
   passport.use(
     new LdapStrategy(
@@ -95,17 +98,24 @@ function enableLdap(config) {
         try {
           const { models } = req;
 
-          const profileUsername = profile.uid || profile.sAMAccountName;
-          const adminRoleFilter = config.get('ldapRoleAdminFilter');
-          const editorRoleFilter = config.get('ldapRoleEditorFilter');
+          if (!profile) {
+            return done(null, false, {
+              message: 'wrong LDAP username or password',
+            });
+          }
 
           // At least with test setup, jpegPhoto is gnarly output
           // Remove prior to logging
           delete profile.jpegPhoto;
-          appLog.debug(profile, 'Found user');
+          appLog.debug(profile, 'Found LDAP profile');
 
-          // not quite sure if uid is returned for ActiveDirectory
+          const profileUsername = profile.uid || profile.sAMAccountName;
+
           if (!profileUsername) {
+            appLog.warn(
+              profile,
+              `Found LDAP profile, but uid or sAMAccountName fields are missing. SQLPad's LDAP auth implementation requires enhancements to support your environment.`
+            );
             return done(null, false, {
               message: 'wrong LDAP username or password',
             });
--- a/server/auth-strategies/ldap.js
+++ b/server/auth-strategies/ldap.js
@@ -13,6 +13,7 @@ function enableLdap(config) {
       {
         passReqToCallback: true,
         integrated: false,
+        // email field from local auth is used for username when using LDAP
         usernameField: 'email',
         passwordField: 'password',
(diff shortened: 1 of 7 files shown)


eladeyal-intel:
I have some (not alot)of expirience with LDAP.

1. Here's an example of how [metabase](http://github.com/metabase/metabase) allows operator to assign LDAP users to different roles. 
https://www.powerupcloud.com/freeipa-integration-with-metabase-part-iii/  You will see that they do not assume that there's a specific attribute (like `employeeType` in your example) but lets one specify a filter which can be arbitrary. For example I would use that to limit access only for a specific group of people in a specific department or mailing list etc. LDAP is used in enterprises so I don't think the person depolying sqlpad necessarily has access to modifying LDAP attributes.

2. Instead of docker+ldap one can test on-line with https://www.forumsys.com/tutorials/integration-how-to/ldap/online-ldap-test-server/ .

rickbergfalk:
@eladeyal-intel for the group mechanism - I'm assuming those queries are done after the user has authenticated? 

What I'm missing from the current existing node implementation is a list of groups the user belongs to. I wonder if that needs to be a second query...

I tried testing using that site, but wasn't able to get it working for some reason.

rickbergfalk:
I should add groups can be used in the existing implementation if the `memberOf` attribute is provided https://ldapwiki.com/wiki/MemberOf. It just isn't available everywhere.

I'll see if I can figure out a query to get groups for a user using an additional library.

eladeyal-intel:
> for the group mechanism - I'm assuming those queries are done after the user has authenticated?

I guess so? Anyway the idea is that the "filter" is much stronger (much more flexible) than anything specific you can come up with (MemberOf... EmployeeType). 


in https://github.com/rickbergfalk/sqlpad/pull/834 @dengc367 wrote "I have tested successfully with the ldap.forumsys.com LDAP Server using their credentials.".

skaravad:
@rickbergfalk @eladeyal-intel my apologies for the delay.  In general, groups are assigned to roles and users are assigned to groups , vs roles directly associated with users as the ldap management becomes cumbersome. 
Majority of the implementations directly work with group membership , so once the user is authenticated, next is authorization, here you can check if the user belongs to a group by using memberOf attribute, from what I see even if the user belongs to a role , the search can still remain the same just that search path would change (at least for openLDAP , not sure if AD does something more advanced)

The current ldap library returns all the groups of the authenticated user but I could not find any reference for searching users in specific groups to make the search more efficient, so I ended up matching env groups to the returned user groups of authenticated user. 

The problem with above is , if multiple user groups need access , then the env var can get really bad. If users need that kind of a granular control, then users must create special ldap group or role imo.

rickbergfalk:
@skaravad I think I know where some of the confusion is with the `memberOf` attribute. For the LDAP server I was testing against, the `memberOf` attribute was not provided in the profile response after log in. LDAP searches using `memberOf` worked though.

@eladeyal-intel I don't doubt that ldap.forumsys.com worked for @dengc367... it is just that at the time I tried it wasn't working for me. I might have messed something up with the configuration.

I've found an additional library to use to run additional queries on ldap. 

@eladeyal-intel Would something like this make more sense?

In config, LDAP search filters would be set per role. For example, it might be something like the following or however fancy someone wants to get:

```sh
SQLPAD_LDAP_ROLE_ADMIN_FILTER="(memberOf=cn=sqlpad-admins,dc=example,dc=com)"
SQLPAD_LDAP_ROLE_EDITOR_FILTER="(memberOf=cn=sqlpad-editors,dc=example,dc=com)"
```

After validating that the user exists, a follow-up query would be run to determine admin/editor. First `(&(memberOf=cn=sqlpad-admins,dc=example,dc=com)(uid={{username}}))` then `(&(memberOf=cn=sqlpad-editors,dc=example,dc=com)(uid={{username}}))`.

If both queries return 0 results, the user will be given the default LDAP role configured. At this point we know the user is allowed to access SQLPad as it matched the initial query filter provided for authentication.

eladeyal-intel:
Looks good for me. Please show (in the Tess or examples) how to use | operator (does it need to be eecaped?)

eladeyal-intel:
Also it's not necessarily uid which matches username . It can be other field such as sAMAaccount (or something like that).

server/auth-strategies/ldap.js
```diff
@@ -28,26 +29,96 @@ function enableLdap(config) {
...
         },
       },
       async function passportLdapStrategyHandler(req, profile, done) {
         try {
           const { models } = req;
-          const mail = profile.mail.toLowerCase();
-          const user = await models.users.findOneByEmail(mail);
-          if (!user) {
+
+          const uid = profile.uid.toLowerCase();
```
dengc367, line 40: This line need change to `const uid = (profile.uid || profile.sAMAccountName).toLowerCase();`, because uid properties in ActiveDirectory is `sAMAccountName`.
When I change this line, the test is successful.

dengc367 (commented):
Great Job.  I test the code successfully when I change the line above reviewed in ActiveDirectory environment. @eladeyal-intel @rickbergfalk . You can merge it to branch `master` after you change that code line.

dengc367 (commented):
For RBAC In ActiveDirectory, `SQLPAD_LDAP_ROLE_ATTRIBUTE` need set to `memberOf` usually.  below is the snippet of ActiveDirectory configuration: 
`
SQLPAD_LDAP_SEARCH_FILTER="(sAMAccountName={{username}})"
SQLPAD_LDAP_ROLE_ADMIN_VALUE="CN=admin,OU=Distribution_Groups,DC=xxx,DC=local"
SQLPAD_LDAP_ROLE_EDITOR_VALUE="CN=editor,OU=Distribution_Groups,DC=xxx,DC=local"
SQLPAD_LDAP_ROLE_ATTRIBUTE="memberOf"
`

dengc367:
PR #858 is created, you can merge is directly from the branch. @rickbergfalk

rickbergfalk:
@eladeyal-intel I'll add the uid/sAMAccountName restriction dynamically based on what comes back from the initial auth check

So the role queries will become `(&(<query for admin membership via config>)(uid=foundProfile.uid))` if profile comes back with uid, `(&(<query for admin membership via config>)(sAMAccountName=foundProfile.sAMAccountName))` if profile comes back with sAMAccountName.

I think the user query can be whatever they want - it can be a single group, multiple groups, include | and &... whatever.

rickbergfalk:
Alright I'm done making modifications. 

Thank you all for initial work LDAP work, design feedback, and being another set of eyes! I really appreciate it!

Issue #741, opened 2020-06-28
"Configure visualization" modal should be draggable


eladeyal-intel:
The "Configure visualization" modal window should be draggable (movable) so that one could inspect the changes in the graph behind it.

rickbergfalk:
What do you think about it being a toggled sidebar similar to schema? I was thinking it could either go on the right hand side, or replace schema sidebar potentially?

eladeyal-intel:
I thought adding a "draggable" decoration would be super easy for one versed in js. 

I'm not sure it's a good idea to put the graphs on the far right and their controlled  on the far left.

rickbergfalk:
Yeah I was thinking the same.

Dragging is possible but I haven't really liked the modal in general. It was kind of a quick and easy solution to removing the previous tabbed design.

Another thing I'm trying to think about and solve is managing query name, tags, sharing, potentially adding description?

Adding a general sidebar that could act as a space for these things might be nice and flexible? Kind of a properties sidebar of sorts.

eladeyal-intel:
In my opinion, the schema window should be of higher priority than other property panels. In my experience it is always visible.  it can be hidden if you are out of screen Real estate. But otherwise it is always visible. It is the same priority and class as the query window and the results window. So that's why I don't think it is interchangeable with the connections list or the visualization configuration.

PR #869, opened 2020-10-30
Move vis modal into sidebar

rickbergfalk:
Moves vis modal into sidebar. This allows the user to see the chart while selecting fields, and also establishes a new pattern/space for expanding on user input.

Eventually tags, sharing query properties can be moved into this sidebar, with a sort of side-tab strip allowing the sidebar to be toggled on and off. 

Note that this is going into the v6 branch so that all UI changes launch at once.

@eladeyal-intel curious what your thoughts since this kind of addresses #741 

image


eladeyal-intel:
👍

Issue #898, opened 2020-11-21
Setting up LDAP authentication


euanmacinnes:
Hi,
   I've enabled the LDAP authentication using environment variables, in the latest docker image (as of 21 Nov 2020), and can't get the LDAP function to work. I don't see any debug information in the docker logs that says that the LDAP is working or not, only one line that states:

```
"Enabling basic authentication strategy."
"Enabling ldap authentication strategy."
"Enabling local authentication strategy."
```

However, upon attempting to log in a (known valid) user's details, there is just an "invalid username / password" message appearing in a pink box on the site, and no additional output in the debug file (i.e. did it find the LDAP server Ok, did it log in Ok, did it find the name, etc..).

Any help would be appreciated. I've configured both PAM authentication with LDAP, and other apps succesfully, using the following settings:

```
          SQLPAD_LDAP_AUTH_ENABLED: 'true'
          SQLPAD_LDAP_URL: 'ldap://192.168.0.101:389'
          SQLPAD_LDAP_SEARCH_BASE: 'ou=Users,dc=amt,dc=com'
          SQLPAD_LDAP_BIND_DN: 'cn=admin,dc=amt,dc=com'
          SQLPAD_LDAP_PASSWORD: '****************'
          SQLPAD_LDAP_SEARCH_FILTER: '(uid={{username}})'
```

This is connecting to an OpenLDAP server that is on the same private LAN. TLS is not enabled on the LDAP.

eladeyal-intel:
1. Don't forget that for LDAP to kick in, the username provided in the login dialog-box must be without "@"

2. I wonder if re-implementing https://github.com/sqlpad/sqlpad/pull/757 again will add more logs. (The LDAP authorization provider was changed after this PR was merged). Unfortunately I'm not sure I am able to test this now myself.

rickbergfalk:
@eladeyal-intel yeah the new provider probably doesn't have the same logging functionality.

@euanmacinnes are your ldap usernames email addresses?

Is there a better way to handle auth if both local and ldap are used? Should there be a separate button? A separate form? Should  both auth mechanisms be tried?

At the very least perhaps if ldap is enabled and local auth is disabled, the input is always sent to the ldap auth strategy regardless of its content.

euanmacinnes:
> @eladeyal-intel yeah the new provider probably doesn't have the same logging functionality.
> 
> @euanmacinnes are your ldap usernames email addresses?
> 
[...]

PR #912, opened 2020-11-30
Fix LDAP auth for profiles without mail attribute

rickbergfalk:
Fixes LDAP auth for case where profile does not have `mail` attribute. In this case the account login will be used in place of email, and stored on `users.email`.  Addresses part of #898.

I have mixed feelings about this as it is kind of abusing the `users` table data model a bit, but it keeps auth flows simple and allows this LDAP case to work without impacting other auth strategies and requiring significant work in the UI/data model/API, and so on.

Two potential changes in the future to address that:
- email could be renamed or officially repurposed as an "external id" field. 
- An additional external id field can be added to `users` table, unique constraint removed from `users.email`, and auth flows, UI, APIs updated to accommodate the new field addition.

### DIFF
--- a/server/auth-strategies/ldap.js
+++ b/server/auth-strategies/ldap.js
@@ -93,13 +93,29 @@ async function enableLdap(config) {
             userIdFilter = `(uid=${profile.uid})`;
           }
 
-          // Email could be multi-valued
-          // For now first is used, but might need to check both in future?
-          let email = Array.isArray(profile.mail)
-            ? profile.mail[0]
-            : profile.mail;
+          // Mail may not always be available on profile
+          // In that case, fall back to user login fields.
+          // This field will go into `users.email` to serve in email's place.
+          // This should probably go into its own field, but this can complicate existing auth and UI
+          // If another field is added in the future, non-emails can be identified and moved accordingly
+          let identifierAttrValue =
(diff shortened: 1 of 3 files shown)


eladeyal-intel:
Is there really a use case for LDAP without email??

I imagine you want the webhooks to use real emails for some (many) use cases.

rickbergfalk:
@eladeyal-intel IDK but it happened in #898 so it is possible

eladeyal-intel:
Yes but #898 can be fixed by updating ldap server configuration. I'm not sure you want to support it in this way. It may create more harm than good.

rickbergfalk:
@eladeyal-intel is the concern with this solution due to a non-email value in the email field or allowing user records without an email address on hand?

I'm considering whether I should add an external id column to SQLPad's users table instead? Email would still need to be made nullable though which could be problematic.

Another thought I had was to make the ldap profile attribute used for email configurable, so that if someone wanted to use uid or sAMAccountName instead it is their decision to do so.

This would allow setups to opt-in to this decision, and would allow SQLPad's non-null email field to remain as is.

eladeyal-intel:
My concern was is with non email data in the email field.

Configable email field may be good idea, and you can also configured the email to be delivered. Is if account name is Jane the mail might be jane@example.com 

I don't have any experience in just making things up here.

PR #917, opened 2020-12-02
Fix clickhouse driver

### LINKED ISSUE #915: Code: 194, e.displayText() = DB::Exception: Invalid authentication: it is not allowed to use X-ClickHouse HTTP headers and other authentication methods simultaneously (version 20.3.9.70 (official build))
Got this error when trying to create connection to clickhouse-server

rickbergfalk:
Fixes needed to support latest version of ClickHouse. This works with latest ClickHouse, but unsure what minimum version is supported. Do we need to add a version indicator and different implementations? 

~Also problematic is the `FORMAT JSON` text fragment. It seems INSERT statements do not like this, so this almost has to be applied for specific statement types?~ JSON is now sent as header

@dengc367 you're more familiar than myself with ClickHouse. Do you have any insight here?

Fixes #915

### DIFF
--- a/server/drivers/clickhouse/_clickhouse.js
+++ b/server/drivers/clickhouse/_clickhouse.js
@@ -28,14 +28,11 @@ function send(config, query) {
   const results = {
     data: [],
   };
-  return fetch(
-    `${config.url}/?user=${config.user}&password=${config.password}&database=${config.database}`,
-    {
-      method: 'POST',
-      body: query,
-      headers: getHeaders(config),
-    }
-  )
+  return fetch(`${config.url}/?database=${config.database}`, {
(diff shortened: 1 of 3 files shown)


rickbergfalk:
@dengc367 I found some examples in ClickHouse docs that show sending format via header. See the Prometheus section where template is specified https://clickhouse.tech/docs/en/interfaces/http/#predefined_http_interface. 

This seems to work, but I don't know when this was added and supported. 

Would you be willing to review?

dengc367:
OK, I will test it soon.

dengc367:
@rickbergfalk  please merge the PR #918 to this branch. I have tested successfully.

PR #922, opened 2020-12-05
Fixes sqlpad/sqlpad#921

tiagosimao:
Added test for HTTP large entities
Added configurable max payload size
Added config documentation

Added `bytes` as a new devDependency to more accurately test body size config parsing

The local build process seems had some side effects, namely adding the `odbc` dependency, a change in the sqlpad bin location and a new package lock. These changes were not intentional and I don't know if they are expected or not, I built locally by typing `sh scripts/build.sh` in the project's root using Node v15.3.0 and MacOS BigSur.


rickbergfalk:
Thanks @tiagosimao this PR looks great! 

I'm a bit unsure about odbc moving to dependencies and the large package-lock.json diff. I'm guessing this is npm 7 behavior. I know they changed some of the peer dependency behaviors and I wouldn't be surprised if optional dependencies changed as well.

Would it be possible to revert the addition of odbc to dependencies? If its problematic with you're setup I'd be comfortable with  merging and figuring out with a follow up too.

Not a huge deal. The reason its optional is more out of convenience to not require all SQLPad installs to support building native add-ons. It might make sense to make it required to prevent potential confusion as to why ODBC is missing from a SQLPad build.

tiagosimao:
Indeed the behavior seems to come from later versions of node/npm.

Reverted and re-ran the local build with node 12 by running
```sh
docker run -v $(pwd)/:/build --workdir /build node:12.20.0-alpine3.10 scripts/build.sh
```

Updated the branch with this less disruptive changes to the dependency management.

rickbergfalk:
Sorry for delay in getting to this - hope to review and merge later tonight

rickbergfalk:
Thanks @tiagosimao!

PR #936, opened 2021-01-29
remove python dependency

eladeyal-intel:


eladeyal-intel:
python2 is EOL. Not sure why you need python anyway, it seems to build and test fine.

rickbergfalk:
Python is needed for building native dependencies in node via node-gyp https://github.com/nodejs/node-gyp#on-unix

Recently the latest sqlite3 prebuilt package was (maybe still is?) not available on s3, causing npm to try to build the dependency natively. I was having trouble getting python3 recognized by tooling to get things to build.

I can see if I can get the python 3 version working.

Should note that python, gcc and make are not included in the final docker image. It's only used during the first stage build.

rickbergfalk:
Got python3 working in #937

PR #939, opened 2021-02-03
Use wrapIntegers to prevent int64 truncation

bruth:
This translates the raw int64 returned over the API and represents
it as a string rather than as a JavaScript integer to prevent
truncating to 52 bits prior to rendering.

Signed-off-by: Byron Ruth <b@devel.io>

### DIFF
--- a/server/drivers/bigquery/index.js
+++ b/server/drivers/bigquery/index.js
@@ -100,7 +100,12 @@ function runQuery(queryString, connection = {}) {
     .createQueryJob(query)
     .then(([job]) => {
       // Wait for the query to finish.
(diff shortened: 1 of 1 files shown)


bruth:
@rickbergfalk Its been a while! This is a pretty small change and I am noticing some of the integration tests with the databases are failing. I've retrieved them a couple times and they don't seem to be consistent where they fail. My change shouldn't have any impact on those tests, so I just wanted to make you aware.

rickbergfalk:
Thanks @bruth! Yeah I've been noticing that. I'm wondering if the current tests running in parallel is maxing out resources or something.

PR #944, opened 2021-02-06
Github action optimizations

rickbergfalk:
FYI @eladeyal-intel tweaking GitHub action things a bit

- Removing Node 13.x from matrix. Odd numbered releases are short-lived, and I think form the basis for what comes of next even numbered release, which are LTS versions of node. So I figure to cut down on test running we can drop 13, but keep 15 since it is the current version https://nodejs.org/en/about/releases/ 
- Uses npm ci for lint install (should be slightly faster and uses lock file for a more reproducible build)
- Uses node 14 for artifact build
- Uses latest SQL Server 2019 image, and increases health check retries. Hoping this cuts down on number of failed tests where action is waiting for server to start up.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/setup-node@v1
         with:
           node-version: ${{ matrix.node-version }}
-      - run: npm install
-      - run: npm install --prefix client
-      - run: npm install --prefix server
+      - run: npm ci
+      - run: npm ci --prefix client
+      - run: npm ci --prefix server
       - run: npm run lint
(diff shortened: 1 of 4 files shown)


eladeyal-intel:
I think "retries" only come into play on transition from healthy to unhealthy, so they won't affect your ci.

rickbergfalk:
I wasn't sure how that affected the container. It seems like SQL Server container just wasn't coming up? Logs have docker ps running repeatedly without SQL Server there. Seeing if adding a container restart option helps it out. Maybe its failing on startup and wasn't retrying? Not sure how to get visibility into that.

rickbergfalk:
Looks like adding a restart policy is helping...

PR #948, opened 2021-02-07
Load schema for initialConnectionId

### LINKED ISSUE #946: Schema not fetched when default connection is set
@rickbergfalk I am not sure exactly where this would be called, (in [`initEditor`](https://github.com/sqlpad/sqlpad/blob/master/client/src/stores/editor-actions.ts#L121)?), but the use case is that I have one connection and it is set to the default so there is way to select or change connections. When a fresh editor loads, the schema doesn't load and there is no way to manually load it. The only side effect I noticed is when I choose a saved query that uses the connection it will load the schema.

I just think what is missing is the schema loading when the default connection is set.

rickbergfalk:
Fixes #946

### DIFF
--- a/client/src/stores/editor-actions.ts
+++ b/client/src/stores/editor-actions.ts
@@ -175,6 +175,9 @@ export const initEditor = async (
 
     const { focusedSessionId } = getState();
     setSession(focusedSessionId, { connectionId: initialConnectionId });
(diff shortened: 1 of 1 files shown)


bruth (approved):
Thanks for the quick fix!

PR #956, opened 2021-02-18
postgres sslmode with self-signed certificates (fixes #954)

### LINKED ISSUE #954: Can't connect to Postgres in SSL mode with self-signed certificate
It's impossible to connect to a Postgres instance via SSL using a self-signed certificate.

When the connection is configured and tested, the "self signed certificate" error appears (passed up from the underlying Postgres driver).

This is desirable because Heroku's free tier [requires SSL mode](https://devcenter.heroku.com/articles/connecting-to-heroku-postgres-databases-from-outside-of-heroku) and only offers self-signed certs.

ianatha:
As per #954:

It's impossible to connect to a Postgres instance via SSL using a self-signed certificate.

When the connection is configured and tested, the "self signed certificate" error appears (passed up from the underlying Postgres driver).

This is desirable because Heroku's free tier requires SSL mode and only offers self-signed certs.

### DIFF
--- a/docs/connections.md
+++ b/docs/connections.md
@@ -152,6 +152,7 @@ A default connection selection can be set using environment variable `SQLPAD_DEF
 | `username`                         | Database Username                                     |   text    |
 | `password`                         | Database Password                                     |   text    |
 | `postgresSsl`                      | Use SSL                                               |  boolean  |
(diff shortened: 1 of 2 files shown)


rickbergfalk:
Thanks @ianatha!

PR #962, opened 2021-02-23
case-insensitive search for /api/queries (fixes #961)

### LINKED ISSUE #961: query search is case-sensitive
1. I save a query called "Inbounds".
2. I search for "inbounds" -- no queries are returned. I expect my "Inbounds" query to be returned.

sqlpad 6.2.1 with Postgres backing store

(This is something I'll look into and make a patch for)

ianatha:


rickbergfalk:
Thanks @ianatha!

PR #989, opened 2021-03-19
Fix ClickHouse FORMAT JSON application and LIMIT syntax parsing

rickbergfalk:
* Prevents `FORMAT JSON` from being added to `INSERT` statements
* Updates `sql-limiter` version that brings support for ClickHouse's `LIMIT <offset_number>,<limit_number>` syntax

FYI @dengc367

### DIFF
--- a/server/drivers/clickhouse/index.js
+++ b/server/drivers/clickhouse/index.js
@@ -49,7 +49,7 @@ async function runQuery(query, connection) {
   // NOTE - sql limiter may need to support ClickHouse's "LIMIT <offset>,<limit>" syntax
   // If a SQLPad users use that syntax, the sql-limiter will use the offset integer as limit integer.
   // This could potentially produce unexpected results
-  const limitedQuery = sqlLimiter.limit(query, ['limit'], maxRowsPlusOne);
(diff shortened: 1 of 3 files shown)


server/drivers/clickhouse/index.js
```diff
@@ -70,7 +70,21 @@ async function runQuery(query, connection) {
...
   });
 
-  rows = await clickhouse.query(limitedQuery + ' FORMAT JSON').toPromise();
+  // clickhouse package will append ' FORMAT JSON' to certain queries,
+  // but does not handle CTEs WITH... SELECT.
+  // This borrows from approach used in package:
+  // https://github.com/TimonKK/clickhouse/blob/9dea3c0c3e4f3e2fe64e59dad762f1db044bf9bf/index.js#L479
+  // This is a bit naive but should cover most cases
+  if (
+    limitedQuery.trim().match(/^(with|select|show|exists)/i) &&
```
dengc367, line 79: ref: Issue #990 , this line means the sql need start with `with|select|show|exists`, but when I comment the first line, the error occurs.

dengc367: It is resolved.

PR #995, opened 2021-03-31
Add support authentication for presto, allow https for presto

maitrungduc1410:
- support authentication for Presto
- Allow user to input `http` or `https` for Presto `host`

### DIFF
--- a/server/drivers/presto/_presto.js
+++ b/server/drivers/presto/_presto.js
@@ -10,7 +10,10 @@ function wait(ms) {
 
 // Get Presto headers from config
 function getHeaders(config) {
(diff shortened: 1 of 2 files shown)


dengc367 (approved):

dengc367:
@maitrungduc1410 the Presto url need the http/https protocol. I add the checkbox to check http/https option in Presto connection form, and fix the Presto url bug. you can check and the test it. #997

maitrungduc1410:
@dengc367 , I think it doesn’t need, user takes responsibility to ensure the hostname is correct, and if any error during connection it’ll print out and the user can know. This provide user flexibility when inputing db host

PR #1031, opened 2021-06-22
odbc support in docker container

JstDust:
This adds ODBC to the main Dockerfile where you can use `--build-arg "ODBC_ENABLED=true"` to build an image with odbc support.
This also adds support for dynamic scripts in `/etc/docker-entrypoint.d` where any sh file is executed before the main sqlpad server allowing others to inject scripts without altering sqlpad entrypoint.

Warning: This changes the base image from alpine to debian buster (node:lts-buster-slim).
This also upgrades to node to lts (node 14) (as recommended by node to use latest LTS for production or in this case the docker images)

The reason to change the base image to `lts-buster-slim` is the common incompatibility with `libc` in `alpine`. Many 3th party odbc providers only provide pre-build binaries see as an example `oracle-odbc-image` and those are generally not compatible with musl-libc. 

NOTE: Do not attempt to provide pre-build docker images for oracle or other 3th party drivers on docker hub you may end in a license hell. And the provided example for oracle is only intended as demo for the changes made to the main Dockerfile

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,18 @@
 # Need to remote into this image and debug some flow? 
 # docker run -it --rm node:12.22.1-alpine3.12 /bin/ash
-FROM node:12.22.1-alpine3.12 AS build
-
-RUN apk add --update --no-cache \
-    python3 \
-    make \
-    g++
-
+FROM node:lts-buster AS build
+ARG ODBC_ENABLED=false
+RUN apt-get update && apt-get install -y \
+    python3 make g++ python3-dev  \
+    && ( \
+        if [ "$ODBC_ENABLED" = "true" ] ; \
+        then \
+         echo "Installing ODBC build dependencies." 1>&2 ;\
+         apt-get install -y unixodbc-dev ;\
+         npm install -g node-gyp ;\
+        fi\
+       ) \
+    && rm -rf /var/lib/apt/lists/*
 RUN npm config set python /usr/bin/python3
 
 WORKDIR /sqlpad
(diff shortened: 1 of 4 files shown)


rickbergfalk (approved):
Thanks for this too! Especially the example and comments.

hrbu:
@rickbergfalk  I'm not that familar with odbc connections. Please could you provide an example for a connection string I can use in the sqlpad dialog: 

grafik

JstDust:
@hrbu I'm not using DSN, i'm using DBQ. DSN is used when you populate /etc/odbc.ini and create data sources that are system static. Using DSN's with sqlpad i.m.o. is a bad choice since it locks you to that one config. Using DBQ instead, allows you to use multiple connections to the same service with different credentials without editing the container each time using DBQ only requires you to alter the setup if your connected systems change. 

So connection string `Driver={Oracle 21 ODBC driver};DBQ=name_of_service_in_tnsnames`

And schema:
```
SELECT OWNER as table_schema, TABLE_NAME as table_name, COLUMN_NAME as column_name, DATA_TYPE as data_type FROM SYS.ALL_TAB_COLUMNS WHERE OWNER != 'SYS' and OWNER != 'SYSTEM' and OWNER != 'AUDSYS'
```
For full documentation on the options please check oracle's documentation, https://docs.oracle.com/en/database/oracle/oracle-database/12.2/adfns/odbc-driver.html#GUID-618B141E-DD46-4907-99C2-486E801CA878

hrbu:
@JstDust So now the connection string and credentials are accepted. But when testing the connection I finally got the following error `[Oracle][ODBC][Ora]ORA-00923: FROM keyword not found where expected` with the following log output
`{"level":50,"time":"2021-09-01T07:33:51.172Z","pid":1,"hostname":"8c836a660d69","name":"sqlpad-app","odbcErrors":[{"state":"HY000","code":923,"message":"[Oracle][ODBC][Ora]ORA-00923: FROM keyword not found where expected\n"}],"stack":"Error: [odbc] Error executing the sql statement","type":"Error","msg":"[odbc] Error executing the sql statement"}`

## My Container Config
grafik

## My SqlPad Test & Result
grafik

JstDust:
@hrbu Sorry for the late response, I just ignore the test, the reason that the test does not work and the reason it will fail with a lot of drivers is because the default test query used against the odbc driver is not compatible with the DB see [source](https://github.com/sqlpad/sqlpad/blob/master/server/drivers/unixodbc/index.js#L79). With the schema query you should be able to just save and list the schema as verification. There is sadly no good test query that would work on all odbc back-ends so no point of loosing sleep over that test. I considered using the schema query as the test but, the size of that query can be extreme, and may also have security related issues (some db's require permissions for schema stuff)

PR #1032, opened 2021-06-28
Added fixes for OIDC illegal redirects.

JstDust:
As commented in the code this fixes redirect issues in the OIDC login flow.

First it makes sure it does not redirect twice when error or user unavailable and does not call `logIn` in those cases anymore.

If you run sqlpad on a fqdn without a subpath (`baseUrl`) chrome and some other browsers did fail on the illegal header value for `location` header. Simply redirecting to `/`  if `baseUrl` is empty, has the desired effect of letting the user properly open sqlpad without refreshing the browser manually.

NOTE: for some unknown reason only on firefox after the illegal redirect is received it did after a few seconds do a redirect to '/'. But i could not find the origin of that redirect event (maybe firefox does this as a recovery feature for invalid response headers???).

### DIFF
--- a/server/routes/auth-oidc.js
+++ b/server/routes/auth-oidc.js
@@ -23,15 +23,24 @@ function handleOidcCallback(req, res, next) {
     if (err) {
       appLog.debug(JSON.stringify(err));
       res.redirect(`${baseUrl}/signin`);
+      return;
     }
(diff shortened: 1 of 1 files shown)


rickbergfalk:
Thanks again @JstDust!

PR #1036, opened 2021-07-15
Render markdown hyperlinks in columns

stephenpascoe:
We need to render hyperlinks in the query table of our sqlpad instance.  Our solution is to detect the standard markdown hyperlink syntax `[description](url)` as a hyperlink.  It would be great if this could be merged upstream to aid our maintenance.

### DIFF
--- a/client/src/common/QueryResultDataTable.tsx
+++ b/client/src/common/QueryResultDataTable.tsx
@@ -47,6 +47,17 @@ const renderValue = (input: any, column: StatementColumn) => {
       </a>
     );
   } else {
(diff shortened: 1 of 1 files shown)


dengc367:
thanks you for your first PR,  @stephenpascoe, can you give us one screenshot of the query result what look like it is.

dengc367 (approved):

stephenpascoe:
Screen Shot 2021-07-24 at 15 54 58

Screenshot shows links to wikipedia generated from a table.  Each cell in the results is hyperlinked.

PR #1037, opened 2021-07-20
Qol/small qol fixes

JstDust:
A bunch of fixes, these were done because i got some user feedback after some use, some are QOL some are UI bug fixes.

- UI BUGs: see "fix search history" & "bug squashing" commits.
- Altered client schema loading to sort schema and tables on name (case insensitive), note: not columns because order can be relevant there
- Fixed issue that all was expanded by default by making a explicit true check on the 'expanded[row.id]', note: it was not consistently always expended or collapsed not sure how that can happen maybe some object pollution with something else then true | false | undefined? (maybe this object should just be a 'Set<string>' instead)
- Altered the schema search to actually search schemas as well if matched tables search is skipped on that schema. and removed schemas that don't yield tables from the table search this helps with dbs with lots of schemas.
- Compression added to the schema's route, this was done because i had a schemas request that was 30-40 mb response, adding the compression resulted in a transfer of only 3.5 mb see commit 'Reduce schema load bottleneck.'.


dengc367 (dismissed):

rickbergfalk:
Thanks for this @JstDust and for reviewing @dengc367!

It seems tests are failing because of package-lock changes. I have been holding off using npm 7 but I suppose its inevitable and I should embrace it. Will see if I can get tests passing before merge.

server/package-lock.json
```diff
@@ -1,8 +1,8441 @@
 {
   "name": "sqlpad",
   "version": "6.7.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
rickbergfalk, line 4: @JstDust Would it be possible to revert this lockfileVersion 2 change and keep it lockfileVersion 1? This might require installing npm 6 temporarily. If this is problematic I can pull into another branch, make changes, then merge.

server/routes/connection-schema.js
```diff
@@ -42,8 +43,17 @@ async function getConnectionSchema(req, res) {
...
 }
 
+// compression is added here becasue a big database server can have huge amount
+// of metadata and since this is not retrieved schema by schema 20mb+ would easily be possible in plain/text
+// on slow connections where a LB does not compress this can be a big bottleneck.
+// better would be to expand this router to provide:
+// - '/api/connections/:connectionId/schema_names' to get all schema names
+// - '/api/connections/:connectionId/schema/:schemaName' to get tables of a schema
+// - '/api/connections/:connectionId/schema/:schemaName/:tableName' to get column information of a schema table
+// and then deprecate the current '/api/connections/:connectionId/schema'
```
rickbergfalk, line 53: While this would be better for smaller API responses, this would not be compatible with autocomplete unfortunately, as autocomplete requires knowledge of the full schema today. Could you either remove this portion of the comment or add a clarification about autocomplete?

rickbergfalk (changes requested):

server/routes/connection-schema.js
JstDust: Hi for now i have removed the comment from the code since adding it there without proper design was not a good plan, sorry for the confusion. 

My though behind that was not alter the clients ability to load the schema. Only the way it is retrieved, currently it is an all or nothing approach this limits the users feedback on what is happening. I do not mind that all is loaded but it should be done with a lot more user feedback. (e.g. a proper loader)
- 'Retrieving schema names....'
- 'Retrieving tables/columns for schema x/x columns retrieved x' << update as data loading promises are completed.
- Show tree

An other advantage of this staged approach is timeout of the client request is less likely to happen due to long wait or abort due to impatient user refreshing the page.

Also the current setup caches the schema flat this is something that could be more structured. Instead of a plain cache it should be considered to load the entire schema into the backend database of sqlpad. Since each driver could provide a different way of loading the data (and be potentially very time consuming) but every driver always presents it the same way. This would allow for searching and loading of schemas to happen against the backend-database that can be indexed. e.g. a tables structure like:
```
connection_schema_version {
 id,
 connection_id,
 timestamp
}
connection_schema (
 connection_schema_version_id,
 schema_name,
 table_name,
 column_name,
 column_type
)
```
The forced refresh can then load a new version and once completed can delete the old version.

server/package-lock.json
JstDust: @rickbergfalk done

rickbergfalk (approved):
Thanks for the changes @JstDust!

Issue #1038, opened 2021-07-21
/api/batches/<batchId> API returns Forbidden


WolfeeLuo:
When I submit a `/api/batches/<batchId>` request following the documentation, a forbidden result was returned.
After viewing the source code, I found the problem is in this line ` if (batch.userId !== user.id) {` .
```
async function batchToReq(req, res, next) {
  try {
    const { models, user, params } = req;
    const batch = await models.batches.findOneById(params.batchId);

    if (!batch) {
      return res.utils.notFound();
    }

    if (batch.userId !== user.id) {
      return res.utils.forbidden();
    }

    req.batch = batch;
    return next();
  } catch (error) {
    return next(error);
  }
}
```
the userId property of batch.userId is read from backend database who's type is varchar, while the id of user.id is the id property of serviceToken object read from database too under the PassportJwtStrategy, but who's type is integer.

I think `userId` of `batched` shoud  be assigned with service_token.name or add a new unique string type column in service_tokens table for stating the belonging of batches/statement/results created via API way.

WolfeeLuo:
well, maybe just altering the type of `id` column of `service_tokens` table from INTEGER to VARCHAR is better.

PR #1042, opened 2021-07-26
Fix api userid data type issue

WolfeeLuo:
#1038 
basically this PR is to fix issue # 1038


rickbergfalk (commented):
Sorry for delay in getting to this @WolfeeLuo! Instead of writing a migration, could we instead cast user.id to string in the comparison with `(batch.userId !== user.id.toString())`?

I'd be okay with the migration if we could somehow add a test with some pre-existing data before this migration runs for each DB.

My concern is mostly with SQLite and Sequelize. I've come across a lot of bugs when altering tables and I've become a bit hesitant to alter tables if we can avoid it.

WolfeeLuo:
Hi @rickbergfalk. Thanks for the reply. 
Actually, type casting with `toString()` method can fix this issue. 
but considering system consistency and elegance, I chose the migration. 
I think `service-token` and `user` should use similar schema as the codes handle API request and user request are not separated.
But, yes....migration may cause some potential issues, especially under mssql because of it's constraints.
If you really concern subsequent issue, I'm OK to close this PR and fix this issue via type casting in code.

rickbergfalk:
@WolfeeLuo I agree that a UUID would be more consistent... I will see if I can write a test for this before I tag next version.

Could you fix lint? I will merge once checks are green :)

WolfeeLuo:
Hi @rickbergfalk , lint has been fixed.
thanks.

rickbergfalk (approved):

rickbergfalk:
Thanks @WolfeeLuo!

PR #1047, opened 2021-08-02
Fix userid comparison issue under API request

WolfeeLuo:


rickbergfalk (approved):
Thanks @WolfeeLuo! Since this comparison doesn't hurt anything, I'm okay merging regardless of what happens with #1042. I think it is still worth trying to make the data model consistent if we can test and ensure there isn't breakage.

PR #1059, opened 2021-09-03
fix: Prevent disclosure of Google OAuth Secret key

y-mehta:
`/api/app` endpoint discloses the value of the `SQLPAD_GOOGLE_CLIENT_SECRET` config variable, and the App endpoint is also accessible to unauthenticated users. Ideally, OAuth Client Secret should not be disclosed publicly (Ref: https://datatracker.ietf.org/doc/html/rfc6819#page-16). 

This PR converts the config value to boolean to prevent inadvertent disclosure.

### DIFF
--- a/client/src/types.ts
+++ b/client/src/types.ts
@@ -117,7 +117,7 @@ export interface AppInfo {
     baseUrl: string;
     defaultConnectionId: string;
     editorWordWrap: boolean;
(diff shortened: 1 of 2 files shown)


rickbergfalk:
🙀 Thanks for catching this!!!

PR #1063, opened 2021-09-06
Added options for increased cookie security

kunagpal:


rickbergfalk:
Thanks @kunagpal!

kunagpal:
@rickbergfalk Not a problem, thanks for reviewing!

pvieito:
Hi! This PR seems to have broken the support with Google Auth when using a HTTPS public URL. Now when trying to log in you end up in a loop, can we revert it?

rickbergfalk:
@pvieito if you set `SQLPAD_SESSION_COOKIE_SAME_SITE = 'Lax'` does that help? From PR https://github.com/sqlpad/sqlpad/pull/1074. 

If possible I'd like to partially revert whichever part breaks the Google Auth. I don't have a setup handy, but I suspect its the strict/Lax that is causing the problem?

pvieito:
@rickbergfalk No, that unfortunately did not solve the problem.

rickbergfalk:
@pvieito Made some of the hardcoded changes in this PR configurable. By default, the values provided to trust proxy and cookie secure are reverted and set to false (the default prior to this PR). The new settings added in #1093 allow for adding these strict configs as desired (FYI @kunagpal)

If either of you try this out in latest and have any issues or feedback please let me know.

PR #1067, opened 2021-09-16
SAML based role management

JstDust:
This allows SAML to handle role management of users in SQLPad.

It uses SAML claims to determine if the user is allowed to be admin.

To not break compatibility `SQLPAD_SAML_ENFORCED_ROLE` must be set to true if you want to use it on each login see `docs/authentication.md` for details.

It also fixes a bug with `samlAuthContext` that was probably caused by a package update.

Most of this is very specific to Azure but should allow other providers to be added with relative ease now.

### DIFF
--- a/server/auth-strategies/saml.js
+++ b/server/auth-strategies/saml.js
@@ -3,6 +3,84 @@ const SamlStrategy = require('passport-saml').Strategy;
 const appLog = require('../lib/app-log');
 
 /**
+ * Lookup table to make code below more readable.
+ * For AzureAD claims see, https://docs.microsoft.com/nl-nl/azure/active-directory/develop/reference-saml-tokens
+ */
+const CLAIMS = {
+  emailaddress:
+    'http://schemas.xmlsoap.org/ws/2005/05/identity/claims/emailaddress',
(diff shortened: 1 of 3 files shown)


server/auth-strategies/saml.js
```diff
@@ -10,7 +88,29 @@ function enableSaml(config) {
...
+     */
+    const samlAdminGroup = config.get('samlAdminGroup');
+
+    /**
+     * It is best to keep this 'editor' and use claims to manage admin role for SQLPAD.
+     * @type string
+     */
+    const samlDefaultRole = config.get('samlDefaultRole');
+
+    // FIX: https://github.com/node-saml/passport-saml/blob/v3.1.2/src/node-saml/types.ts#L101
```
rickbergfalk, line 106: 🙏 thanks for catching this!

rickbergfalk (approved):
Thanks @JstDust this looks great! Thanks a bunch for adding comments and documentation while you're at it 💪 

I'm not familiar with SAML or Azure AD implementation of it, but would any of this role work be usable by other SAML providers?

Regardless, I'll merge for the bugfix and enhanced functionality. If other SAML setups can be given role management support that can be addressed at a later time.

PR #1072, opened 2021-10-18
Limit listed connection based on user access.

JstDust:
The current UI pulls all connections regardless of access this changes that by only returning connections that the user has access to.
Condition to get connections.
- user has role admin, no access checks are executed and you get all connections.
- user must have a connection exposed through access menu by an admin.
  - The user access list is retrieved by using the current `user.id` or the magic `consts.EVERYONE_ID`.
  - If the user's access list contains `consts.EVERY_CONNECTION_ID` all connections are returned.
  - Otherwise the `connection.id` must be contained in the user access list

### DIFF
--- a/server/models/connection-accesses.js
+++ b/server/models/connection-accesses.js
@@ -85,6 +85,15 @@ class ConnectionAccesses {
     return this.sequelizeDb.ConnectionAccesses.findOne({ where });
   }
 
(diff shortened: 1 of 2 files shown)


rickbergfalk (approved):

rickbergfalk:
Thanks @JstDust!

PR #1074, opened 2021-10-20
Allow setting session cookie SameSite.

JstDust:
This will allow people to fix issues with login routing that is dependent on sessions and redirects. e.g. firefox with oidc. Chrome apparently is not following 'strict' mode to the letter.


rickbergfalk:
Thanks @JstDust!

PR #1076, opened 2021-10-26
Update variable name for region

allinux:
* Fix variable name for region (region => awsRegion)


rickbergfalk:
Thanks for finding this @allinux!

PR #1080, opened 2021-11-04
Add TiDB support

hooopo:
[TiDB](https://github.com/pingcap/tidb) is a MySQL-compatible database that supports Hybrid Transactional and Analytical Processing (HTAP) workloads. It can be used as a replacement for MySQL (same protocol and syntax). We have already seen some TiDB users using TiDB with SQLPad together. So it would be great that we can have TiDB in the supported  databases list.

This is the result of my verification using sqlpad and TiDB on my local server:

image


rickbergfalk:
Thanks @hooopo !

PR #1084, opened 2021-11-19
Update index.js

ted46045:
Updates schema query to get foreign tables too. Use case is using steampipe(.io) in service mode, all tables are foreign tables.


rickbergfalk:
Thanks @ted46045!

PR #1088, opened 2021-11-23
[Athena] Added support for reading the schema from Glue

pvieito:
Added support for reading the schema from Glue, which includes views info in addition to tables info.


rickbergfalk (approved):

rickbergfalk:
Thanks @pvieito!

PR #1090, opened 2021-12-14
Allow base64 for mysql ssl

JstDust:
It's a simple to allow adding CA, Cert and Key to the connection config of SQLPad in base64 form, this prevents the need of storing files locally, and enabling container deployments where persistent volumes are not an option to work with mysql instances where SSL is enforced.

To use it prefix the path field with `base64:` and add the base64 encoded content of the file in the path field to have it provide a base64 decoded buffer instead of buffer from `fs.readFileSync`.

### DIFF
--- a/server/drivers/mysql/index.js
+++ b/server/drivers/mysql/index.js
@@ -34,6 +34,18 @@ function getSchemaSql(database) {
   `;
 }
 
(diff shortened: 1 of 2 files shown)


rickbergfalk:
Thanks @JstDust!

PR #1097, opened 2021-12-30
helm chart for sqlpad

linjinbao666:
install and manage sqlpad using helm in one command


rickbergfalk:
Thanks for this @linjinbao666! Apologies in delay in getting to it.

PR #1101, opened 2022-02-08
Use query timeout from config for postgres drivers

Uriopass:
Currently, there is no way to limit the query timeout, which can lead to very long running queries.
This can severly impact the database as autovacuums cannot delete newly produced rows while a transaction is running. (xmin)

I used the same code as the BigQuery driver.

the query_timeout param is documented here:
https://node-postgres.com/api/client

### DIFF
--- a/server/drivers/postgres/index.js
+++ b/server/drivers/postgres/index.js
@@ -31,6 +31,11 @@ class Client {
       stream: createSocksConnection(connection),
     };
 
+    const queryTimeout = parseInt(connection.queryTimeout, 10);
+    if (queryTimeout) {
+      pgConfig.query_timeout = queryTimeout * 1000;
+    }
+
     // TODO cache key/cert values
     if (connection.postgresKey && connection.postgresCert) {
       pgConfig.ssl = {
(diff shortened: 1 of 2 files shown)


server/drivers/postgres/index.js
```diff
@@ -29,6 +30,7 @@ class Client {
       port: connection.port || undefined,
       ssl: connection.postgresSsl,
       stream: createSocksConnection(connection),
+      query_timeout: getTimeoutSeconds() * 1000,
```
rickbergfalk, line 33: Instead of sourcing this from app config, can this instead be added to postgres fields? This would allow setting different timeouts for different connections, and have it exposed via UI.

To do this, you can add the following to the `fields` array at bottom of the file:
```
{
    key: 'queryTimeout',
    formType: 'TEXT',
    label: 'Query Timeout (seconds)',
  }
  ```
  
and then access that value when creating the connection with `connection.queryTimeout`

rickbergfalk (changes requested):
Thanks for this submission! I'm open to this addition but was wondering if it could be changed to use connection config instead of application config

server/drivers/postgres/index.js
Uriopass: Yes I'm just wondering how to connect it to the env variable SQLPAD_TIMEOUT_SECONDS. Which is the simplest to change.

Uriopass: We want to enforce this timeout so that even "malicious users" (or junior developers of our infrastructure) cannot make too long running queries

Uriopass: Ok I think I get it

Uriopass:
Patched

rickbergfalk:
Thanks @Uriopass !

PR #1106, opened 2022-03-07
Default roles for Google Auth

kunagpal:
Use case: Reduce manual admin overhead through configurable default roles.

This brings Google OAuth behaviour closer to parity with that of [existing auth strategies like LDAP](https://github.com/sqlpad/sqlpad/blob/master/server/auth-strategies/ldap.js#L17).


rickbergfalk:
Thanks for this @kunagpal! Looks great

PR #1116, opened 2022-03-23
Option to overwrite default max row numbers

Bermpje:
We have a use case where specific users are allowed to retrieve a larger amount of rows than allowed by default. To so so, we created an extra maxrows_override field for the following drivers:

- athena
- mysql
- mysql2
- postgresql
- sqlserver
- snowflake
- unixodbc


server/drivers/athena/index.js
```diff
@@ -268,8 +276,29 @@ const fields = [
...
+ * @returns
+ */
+function resolvePositiveNumber(num, defaultValue) {
+  if (num == null) return defaultValue;
+  if (typeof num === 'string') num = Number.parseInt(num, 10);
+  if (typeof num !== 'number') return defaultValue;
+  if (!Number.isFinite(num)) return defaultValue;
+  if (num > 0) return num;
+  return defaultValue;
+}
```
rickbergfalk, line 300: Could this be extracted and put in util/lib directory with some tests?

rickbergfalk (changes requested):
Thanks for contribution @Bermpje!

Could the `resolvePositiveNumber` be moved to its own file in `server/lib` and have a few test cases written for it?

server/drivers/athena/index.js
Bermpje: Done!

rickbergfalk (approved):
Looks great! Thanks @Bermpje!

PR #1129, opened 2022-05-28
feat: add TDengine database support

arielyang:
TDengine is  a High-Performance, Scalable, Time-Series Database with SQL support

Website: https://tdengine.com

### DIFF
--- a/server/drivers/index.js
+++ b/server/drivers/index.js
@@ -20,6 +20,7 @@ const drivers = {
   sqlserver: require('./sqlserver'),
   trino: require('./trino'),
   vertica: require('./vertica'),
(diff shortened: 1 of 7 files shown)


rickbergfalk:
Thanks @arielyang! Would be happy to pull this in.

I've just merged a change to use yarn as package manager instead of npm. Can you pull those in and address the conflicts?

Apologies for the socket-security alert. It is a false positive and I'm not sure how to go about flagging it as such yet with socket.

rickbergfalk:
@arielyang thanks for the effort on this but I'm going to close this one out. SQLPad development is winding down and I'd like to avoid taking more functionality on.

PR #1138, opened 2022-07-05
Support base URL overrides

masch712:
For when UI and API are hosted at different URL paths, for example serving the built HTML/CSS/JS static content from a CDN, but serving the API from behind some API gateway.

I'd be happy to add docs if maintainer(s) like this feature.

### DIFF
--- a/client/src/utilities/api.ts
+++ b/client/src/utilities/api.ts
@@ -302,7 +302,7 @@ export const api = {
           config: {
             ...fetched.config,
             baseUrl:
-              import.meta.env.VITE_UI_BASE_URL_OVERRIDE ||
+              import.meta.env.VITE_SPA_BASE_URL_OVERRIDE ||
               fetched.config.baseUrl,
           },
         };
--- a/client/src/utilities/api.ts
+++ b/client/src/utilities/api.ts
@@ -19,7 +19,8 @@ import {
   UserSelfUpdate,
 } from '../types';
 import { Links, parseLinkHeader } from '../utilities/parse-link-header';
-import baseUrl from './baseUrl';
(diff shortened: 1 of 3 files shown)


rickbergfalk:
Thanks for the contribution @masch712! I'd be happy to pull this in if you can add some documentation? Maybe a short new section in `docs/configuration.md` since this config is kind of a special case. Doesn't need to be much. A short blurb on why someone would set this like you have here in PR description would be fine.

masch712:
Thanks @rickbergfalk !  I left some details in `docs/configuration.md`.  After further testing, and in the process of writing the docs, I realized I oversold this feature in my initial PR description :-).  This feature doesn't really support hosting the static content in a CDN, it's more for a Backend-For-Frontend situation where the BFF handles user authentication and serves the sqlpad single-page app, but the API is hosted in a separate domain (behind an API Gateway, for example, which enforces authorization) and the BFF relays requests to the API.  Again, sort of a custom situation for my own setup, but this code change works for me.

rickbergfalk (approved):

rickbergfalk:
Hey @masch712 - Wanted to give you a heads up that I merged but then had to revert this. The dev flow of using the `config.dev.env` was no longer allowing user login due to the changes.

Would happily merge again once that is solved for.

masch712:
@rickbergfalk yikes, sorry about the breaking chnage!  Fix is straightfoward; I created a new PR for it: https://github.com/sqlpad/sqlpad/pull/1141

rickbergfalk:
No worries! It didn't go out in a release or anything and I should have caught it in review :)

PR #1145, opened 2022-07-24
Support base URL overrides

masch712:
* Support base URL overrides

For when UI and API are hosted at different URL paths.

* Add documentation

* Should be fully backwards compatible (i.e. running `npm run build` in sqlpad/client directory should pass without error)

Co-authored-by: Aaron Asch <aasch@athenahealth.com>


masch712:
@rickbergfalk  I think I've resolved conflicts and fixed the non-backwards-compatibility bug, but I need to test still.  I"ll let you know when this is ready.

masch712:
@rickbergfalk all right, ready for review!

rickbergfalk:
Thanks again @masch712!

PR #1203, opened 2023-10-03
chore: Add .qovery.yml file to deploy app with Qovery 🚀

cillvius:


rickbergfalk:
Appreciate the PR @cillvius but I'm going to close this out. Since the project is in maintenance mode we're looking to decrease support/functionality instead of expand.

PR #1234, opened 2024-05-29
replaces connection dropdown with react-select using autocomplete

ocomes:
In use cases with to many connections it not usable the dropdown to find the correct one. For this reason i've added react-select to use the autocomplete feature


ocomes:
Hi @rickbergfalk,
the security check failed due to Socket Security Issue.
Cant you check and validate my pr? 
thanks

rickbergfalk:
Appreciate your effort with your PR @ocomes! Unfortunately we're not looking to make any changes like this, as it introduces more maintenance burden with the addition of dependencies.

If this project had more of a future, I think making this change would be the right direction though. I encourage you to maintain a fork of SQLPad!

